PR #46, opened 2018-02-27
Replace os-locale with native electron api, app.getLocale

Hinton:
The native electron API seems to grab a more *correct* locale than the `os-locale` package on my machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -223,7 +223,6 @@
     "electron-updater": "2.20.1",
     "keytar": "4.1.0",
     "node-forge": "0.7.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -223,7 +223,6 @@
     "electron-updater": "2.20.1",
     "keytar": "4.1.0",
     "node-forge": "0.7.1",
-    "os-locale": "2.1.0",
```
kspearrin: There's another package.json in src that needs this removed from as well.

Hinton: Done.

PR #75, opened 2018-04-10
 Add language selection in settings

CostantiniMatteo:
I've added an option to choose the language to be used instead of the system language and a basic description in English.

### DIFF
--- a/src/app/accounts/settings.component.ts
+++ b/src/app/accounts/settings.component.ts
@@ -15,6 +15,8 @@ import { StorageService } from 'jslib/abstractions/storage.service';
 
 import { ConstantsService } from 'jslib/services/constants.service';
 
+import { SupportedLocalesLanguage } from '../../services/i18n.service';
+
 @Component({
     selector: 'app-settings',
     templateUrl: 'settings.component.html',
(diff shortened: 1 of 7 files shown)


kspearrin:
Thanks. This looks pretty good. I will look into merging it soon.

src/services/i18n.service.ts
```diff
@@ -4,11 +4,15 @@ import * as path from 'path';
 import { I18nService as I18nServiceAbstraction } from 'jslib/abstractions/i18n.service';
 
 // First locale is the default (English)
-const SupportedTranslationLocales = [
-    'en', 'cs', 'da', 'de', 'es', 'et', 'fi', 'fr', 'hr', 'hu', 'id', 'it', 'ja',
-    'nb', 'nl', 'pl', 'pt-BR', 'pt-PT', 'ro', 'ru', 'sk', 'sv', 'tr', 'uk', 'vi',
-    'zh-CN', 'zh-TW',
-];
+export const SupportedLocalesLanguage = {
```
kspearrin: Can you revert this back to a standard string array? We will need to support localized strings for the language names.

src/app/accounts/settings.component.ts
```diff
@@ -43,11 +47,17 @@ export class SettingsComponent implements OnInit {
             { name: i18nService.t('onRestart'), value: -1 },
             { name: i18nService.t('never'), value: null },
         ];
+
+        this.locales = [ { locale: null, language: 'Default' } ];
+        Object.keys(SupportedLocalesLanguage).forEach((i) => {
+            this.locales.push({locale: i, language: (SupportedLocalesLanguage as any)[i]});
```
kspearrin: there here you pull the language name with `i18nService.t(localId)`...

kspearrin:
Thanks!

PR #130, opened 2018-08-29
when invoke ctrl/command+f select all text in the "Search Vault" input

ra1nb0w:
thanks to @gmittone


src/app/vault/vault.component.ts
```diff
@@ -98,7 +98,9 @@ export class VaultComponent implements OnInit, OnDestroy {
                         await this.addFolder();
                         break;
                     case 'focusSearch':
-                        (document.querySelector('#search') as HTMLInputElement).focus();
+                        let search = document.querySelector('#search') as HTMLInputElement;
+                        search.focus();
+                        search.setSelectionRange(0, (search.value || '').length);
```
kspearrin: Why not just `.select()`?

ra1nb0w: Because we don't thought about it :(
Tested and works correctly. Do you need a new commit or do yourself?
thanks

PR #150, opened 2018-10-20
Add keybinding to open settings

arunvelsriram:
This PR adds a keyboard shortcut to open settings.

In OS X: `Cmd + ,`
In Windows/Linux: `Ctrl + ,` (not tested)

### DIFF
--- a/src/main/menu.main.ts
+++ b/src/main/menu.main.ts
@@ -370,6 +370,7 @@ export class MenuMain extends BaseMenu {
                 label: this.main.i18nService.t('settings'),
                 id: 'settings',
                 click: () => this.main.messagingService.send('openSettings'),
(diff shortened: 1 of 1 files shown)


kspearrin:
Thanks for the PR. A couple of questions.

1. What is the primary use case you are seeing for needing to open settings with a shortcut?
2. Why was `Cmd + ,` chosen? Is this some standard on macOS?

arunvelsriram:
2. Why was Cmd + , chosen? Is this some standard on macOS?

Its the standard for desktop applications in Mac OS

1. What is the primary use case you are seeing for needing to open settings with a shortcut?

Being used to this standard I tried to open settings using the shorcut and found that it didn't work.

PR #168, opened 2018-11-25
Close to tray

h44z:
Allow the app to be minimized to the tray icon if the window gets closed. This is disabled by default.

Tested (and therefore enabled for linux) on Archlinux.

Needed jslib changes: https://github.com/bitwarden/jslib/pull/21

### DIFF
--- a/src/app/accounts/settings.component.ts
+++ b/src/app/accounts/settings.component.ts
@@ -30,6 +30,7 @@ export class SettingsComponent implements OnInit {
     disableGa: boolean = false;
     disableFavicons: boolean = false;
     enableMinToTray: boolean = false;
+    enableCloseToTray: boolean = false;
     enableTray: boolean = false;
     showMinToTray: boolean = false;
     locale: string;
@@ -78,6 +79,7 @@ export class SettingsComponent implements OnInit {
         this.lockOption = await this.storageService.get<number>(ConstantsService.lockOptionKey);
         this.disableFavicons = await this.storageService.get<boolean>(ConstantsService.disableFaviconKey);
         this.enableMinToTray = await this.storageService.get<boolean>(ElectronConstants.enableMinimizeToTrayKey);
+        this.enableCloseToTray = await this.storageService.get<boolean>(ElectronConstants.enableCloseToTrayKey);
         this.enableTray = await this.storageService.get<boolean>(ElectronConstants.enableTrayKey);
         this.locale = await this.storageService.get<string>(ConstantsService.localeKey);
         this.theme = await this.storageService.get<string>(ConstantsService.themeKey);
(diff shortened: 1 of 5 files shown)


src/locales/de/messages.json
```diff
@@ -806,6 +806,12 @@
   "enableMinToTrayDesc": {
     "message": "Zeige ein Symbol in der Taskleiste, wenn das Fenster minimiert wird."
   },
+  "enableCloseToTray": {
```
kspearrin: revert this file. only English is added during development and translations are done via crowdin.

src/main/menu.main.ts
```diff
@@ -365,6 +365,10 @@ export class MenuMain extends BaseMenu {
                 click: () => this.main.messagingService.send('lockVault'),
                 accelerator: 'CmdOrCtrl+L',
             },
+            {
+                label: this.i18nService.t('quitBitwarden'),
+                role: 'quit',
```
kspearrin: This is going to result in two quit menu options on mac. See https://github.com/bitwarden/jslib/blob/master/src/electron/baseMenu.ts#L121

src/app/accounts/settings.component.ts
```diff
@@ -74,10 +75,11 @@ export class SettingsComponent implements OnInit {
     }
 
     async ngOnInit() {
-        this.showMinToTray = this.platformUtilsService.getDevice() === DeviceType.WindowsDesktop;
+        this.showMinToTray = this.platformUtilsService.getDevice() === DeviceType.WindowsDesktop || this.platformUtilsService.getDevice() === DeviceType.LinuxDesktop;
```
kspearrin: We sure this works well on Linux? Seems that it was explicitly set to windows for a reason.

h44z: I have only tested this on Archlinux where it works fine

src/main/menu.main.ts
h44z: @kspearrin Hm I see. But as a Windows/Linux user it would make more sense if the close button is located in the "File" menu I guess. What do you think?

kspearrin: I don't have a problem with it being on the File menu for Windows and Linux. We just dont want it duplicated on macOS .

src/app/accounts/settings.component.ts
h44z: Just tested it under ubuntu 18.10, seems to have a bug where it duplicates the tray icon each time the window gets minimized/closed. I have to investigate this.

src/main/menu.main.ts
h44z: @kspearrin I have now enabled the menu only for windows and linux

src/app/accounts/settings.component.ts
h44z: @kspearrin I have now reverted the change. So the tray functionality is only available on Windows.

h44z:
@kspearrin are the code changes now ok?

BrMBr:
Not closing to tray on Linux.

h44z:
This is currently not enabled for Linux :(

PR #169, opened 2018-11-25
Allow app to start minimized (as tray icon)

h44z:
If this setting is enabled, the main window will be hidden after startup. Disabled by default.

This is useful if Bitwarden gets started automatically after boot.

Needed jslib changes: https://github.com/bitwarden/jslib/pull/22


h44z:
@kspearrin any thoughts on this?

kspearrin:
How is Bitwarden getting started automatically after boot? I don't think that is supported at the moment.

h44z:
On Windows: simply put a link to bitwarden.exe to your autostart folder

BrMBr:
It needs to start minimized on Linux as well.

PR #183, opened 2019-01-02
Disable text selection on non-editable UI elements

ghost:
Previously any text could be selected. This PR allows only text in fields the user can edit to be selected.

### DIFF
--- a/src/scss/base.scss
+++ b/src/scss/base.scss
@@ -11,6 +11,16 @@ html, body {
     font-family: $font-family-sans-serif;
     font-size: $font-size-base;
     line-height: $line-height-base;
+    user-select: none;
+}
+
+.row-main,
+.box-content-row {
+    user-select: text;
+}
+
+.row-label {
+    user-select: none;
 }
 
 body {


kspearrin:
What issue does this cause? I'm thinking there may be cases where a user would want to select the label. Custom fields perhaps?

Frizbunny:
Personally I haven't been able to come up with a situation where selecting the label is useful, and as it is now Cmd + A selects everything in the window (sidebar, other items, etc.) instead of just the information in important fields. By disabling selection on everything else it means that a quick "select all" grabs the information the user was likely intending to select.

Also, since most native apps don't allow you to select all text, it makes the desktop client feel a little more at home... or I'm just a strange one for trying to select text in every application 😅

lukescammell:
I use label selection every time I create a more complex login form, due to a lack of being able to create custom form templates. So right now, I would rather be able to copy the field labels.

I could support this if the labels also got a simple "copy label" button, much like the forms themselves, but I imagine that would end up looking very messy.

eliykat:
I'm closing this because the CLA has not been signed and the user has now deleted their account.

PR #216, opened 2019-02-19
Added copy username and copy password shortcuts

rafadc:
The purpose of the PR is to make possible to copy a password and a username from the application without relying in your mouse. At this very moment I still have to select the cipher manually but this is a step in that direction. It is useful on its own since it avoids one click. If I am not doing anything terrible I'd like to continue adding shortcuts so at least basic functionality is available with no need to reach the mouse. 

image

Since ideally this should be executable from anywhere I added it to the edit menu. 

image

I had to do some changes in jslib so it depends on https://github.com/bitwarden/jslib/pull/29

This is my first PR so I may be wrong in so many things. Please, ask for whatever you consider necessary. I am not familiar with the codebase so I may be placing some things in the incorrect place or doing many incorrect stuff. 

- [x] : Initial proof of concept
- [x] : Avoid to change jslib

### DIFF
--- a/src/app/vault/vault.component.ts
+++ b/src/app/vault/vault.component.ts
@@ -103,12 +103,13 @@ export class VaultComponent implements OnInit, OnDestroy {
                         await this.addFolder();
                         break;
                     case 'copyUsername':
-                        if (this.cipherId != null && this.selectedCipher().login.username != null) {
+                        if (this.selectedCipher() != null && this.selectedCipher().login.username != null) {
                             this.copyValue(this.selectedCipher().login.username, 'username');
                         }
                         break;
                     case 'copyPassword':
-                        if (this.cipherId != null && this.selectedCipher().login.password != null) {
+                        console.log(this.selectedCipher());
+                        if (this.selectedCipher() != null && this.selectedCipher().login.password != null) {
                             this.copyValue(this.selectedCipher().login.password, 'password');
                         }
                         break;
@@ -621,6 +622,10 @@ export class VaultComponent implements OnInit, OnDestroy {
     }
 
     private selectedCipher() {
-        return this.ciphersComponent.ciphers.find(cipher => cipher.id == this.cipherId);
+        if (this.cipherId == null) {
+            return null;
+        } else {
+            return this.ciphersComponent.ciphers.find(cipher => cipher.id == this.cipherId);
+        }
     }
 }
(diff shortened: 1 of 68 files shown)


src/app/vault/vault.component.ts
```diff
@@ -609,4 +619,8 @@ export class VaultComponent implements OnInit, OnDestroy {
         this.addOrganizationId = null;
         this.addCollectionIds = null;
     }
+
+    private selectedCipher() {
```
rafadc: Is there an easiest way to access the currently selected cipher?

kspearrin: Seems ok, but this can return null, which you don't account for when calling this function

src/app/vault/vault.component.ts
```diff
@@ -246,12 +254,14 @@ export class VaultComponent implements OnInit, OnDestroy {
                 if (cipher.login.username != null) {
                     menu.append(new remote.MenuItem({
                         label: this.i18nService.t('copyUsername'),
+                        accelerator: 'CmdOrCtrl+shift+U',
```
kspearrin: Is a shortcut necessary on this menu too? Not sure if it does anything here.

rafadc: I really just put it here so it appears when you right click on the context menu. Just as information. If you feel like it I can remove it.

src/app/vault/vault.component.ts
rafadc: Changed this.

src/app/vault/vault.component.ts
```diff
@@ -102,9 +102,18 @@ export class VaultComponent implements OnInit, OnDestroy {
                     case 'newFolder':
                         await this.addFolder();
                         break;
+                    case 'copyUsername':
+                        if (this.selectedCipher() != null && this.selectedCipher().login.username != null) {
+                            this.copyValue(this.selectedCipher().login.username, 'username');
+                        }
+                        break;
+                    case 'copyPassword':
+                        console.log(this.selectedCipher());
```
nea: The console.log() is probably not supposed to stay or is it?

rafadc: Yes, I forgot to push this change.

rafadc:
I did all the requested changes ;)

Thanks for the feedback.

lukescammell:
Will this work on Windows and Linux using `Ctrl+Shift+U` and `Ctrl+Shift+P`?

macav:
is there any chance the conflicts could get resolved and the PR approved? This would be a really nice addition to the app!

lobscouse:
Agree, I actually just implemented this in the same way, though this is my first electron project, and I can't get it to build, so not bothering to submit PR, but would sorely like at least this functionality.

elias-pap:
Hello, the merged PR #379 seems a duplicate. This can possibly be closed.

PR #237, opened 2019-03-26
Drag n drop sorting for custom fields

Kovah:
This pull request implements the ability to sort custom fields via drag n drop while editing them. The order is preserved when saving the entry.
The drag n drop feature uses the Angular CDK DragDropModule to provide frictionless support.

Preview:
bitwarden_sorting_preview

I had to move the styles for the `.box-content-row` class out of the `.box` declaration to prevent hundreds of lines code duplication. I hope this won't be a problem but I couldn't find any conflicting styles.

:warning: This pull request has a dependent pull request in the bitwarden/jslib repository which needs to be merged first before this pull request can be merged. Once it's merged I will update the submodule reference.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -413,6 +413,15 @@
         "tslib": "^1.9.0"
       }
     },
+    "@angular/cdk": {
+      "version": "7.3.6",
+      "resolved": "https://registry.npmjs.org/@angular/cdk/-/cdk-7.3.6.tgz",
+      "integrity": "sha512-ZuOz8nQk0bdo8YyNFcwnmSl4MPaQDAFTbLK29w4Vd/LfPnhBI3pAr0wVuPFb0fl3eSvvUrfTb/+kPbQcE07A0A==",
+      "requires": {
+        "parse5": "^5.0.0",
+        "tslib": "^1.7.1"
+      }
+    },
     "@angular/common": {
       "version": "7.2.1",
       "resolved": "https://registry.npmjs.org/@angular/common/-/common-7.2.1.tgz",
@@ -1397,12 +1406,20 @@
       "dev": true
     },
     "async": {
-      "version": "2.6.1",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.1.tgz",
-      "integrity": "sha512-fNEiL2+AZt6AlAw/29Cr0UDe4sRAHCpEHh54WMz+Bb7QfNcFw4h3loofyJpLeQs4Yx7yuqu/2dLgM5hKOs6HlQ==",
+      "version": "2.6.2",
+      "resolved": "https://registry.npmjs.org/async/-/async-2.6.2.tgz",
+      "integrity": "sha512-H1qVYh1MYhEEFLsP97cVKqCGo7KfCyTt6uEWqsTBr9SO84oK9Uwbyd/yCW+6rKJLHksBNUVWZDAjfS+Ccx0Bbg==",
       "dev": true,
       "requires": {
-        "lodash": "^4.17.10"
+        "lodash": "^4.17.11"
+      },
+      "dependencies": {
+        "lodash": {
+          "version": "4.17.11",
+          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
+          "integrity": "sha512-cQKh8igo5QUhZ7lg38DYWAxMvjSAKG0A8wGSVimP07SIUEK2UO+arSRKbRZWtelMtN5V0Hkwh5ryOto/SshYIg==",
+          "dev": true
+        }
       }
     },
     "async-done": {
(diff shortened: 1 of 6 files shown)


kspearrin:
I am still confused about why there are so many style changes.

src/app/vault/add-edit.component.html
```diff
@@ -261,36 +261,42 @@
...
-                                        [ngClass]="{'fa-eye': !f.showValue, 'fa-eye-slash': f.showValue}"></i>
+                    <div cdkDropList (cdkDropListDropped)="drop($event)">
+                        <ng-container *ngIf="cipher.hasFields">
+                            <div class="box-content-row box-content-row-multi box-draggable-row" cdkDrag
+                                *ngFor="let f of cipher.fields; let i = index; trackBy:trackByFunction"
+                                [ngClass]="{'box-content-row-checkbox': f.type === fieldType.Boolean}">
+                                <span class="box-drag-handle" title="{{'drag' | i18n}}" cdkDragHandle>
+                                    <i class="fa fa-bars"></i>
+                                </span>
+                                <a href="#" class="box-content-delete" appStopClick (click)="removeField(f)"
```
kspearrin: i dont see this `box-content-delete` class being used anywhere. why was it added?

src/app/vault/add-edit.component.html
```diff
@@ -261,36 +261,42 @@
...
-                            </div>
-                            <input id="fieldValue{{i}}" name="Field.Value{{i}}" type="checkbox" [(ngModel)]="f.value"
-                                *ngIf="f.type === fieldType.Boolean" appTrueFalseValue trueValue="true"
-                                falseValue="false">
-                            <div class="action-buttons" *ngIf="f.type === fieldType.Hidden">
-                                <a class="row-btn" href="#" appStopClick appBlurClick
-                                    title="{{'toggleVisibility' | i18n}}" (click)="toggleFieldValue(f)">
-                                    <i class="fa fa-lg"
-                                        [ngClass]="{'fa-eye': !f.showValue, 'fa-eye-slash': f.showValue}"></i>
+                    <div cdkDropList (cdkDropListDropped)="drop($event)">
```
kspearrin: Can you explain what changed in this block? Having a hard time seeing the diff here. Looks like a new wrapping div and the `box-drag-handle` span. Is that all?

src/scss/box.scss
```diff
@@ -1,6 +1,7 @@
 @import "variables.scss";
```
kspearrin: confused about why so much change

src/app/vault/add-edit.component.html
```diff
@@ -261,36 +261,42 @@
...
+                        <ng-container *ngIf="cipher.hasFields">
+                            <div class="box-content-row box-content-row-multi box-draggable-row" cdkDrag
+                                *ngFor="let f of cipher.fields; let i = index; trackBy:trackByFunction"
+                                [ngClass]="{'box-content-row-checkbox': f.type === fieldType.Boolean}">
+                                <span class="box-drag-handle" title="{{'drag' | i18n}}" cdkDragHandle>
```
kspearrin: does a locale entry for `drag` exist?

Kovah:
Here's an overview on the style and HTML changes:

* Wrapped the `ng-container` with a div that handles the cdkDropList, to keep the "Add Field" box out of the sorting list.
* Added a span that works as the drag handle. (Will remove the box-content-delete, used it for styling but works without it)
* Moved the whole styles for `.box-content-row` out of the `.box` block (maybe the reason for all those changes in the styles file), but there are no style changes for the box-content-row itself.
* Added a new `.box-drag-handle` block for styling the drag handle.
* Added a new `.cdk-drag-preview` block for styling the moving custom field box.

A locale entry for `drag` does not exist yet, will add it in a minute.

kspearrin:
Thanks for the details. But, I still don't understand why `.box-content-row` needs to be outside of `.box` now. Is it because the element shown when dragging it around is outside of that scope?

Kovah:
Correct. The dragged field is copied/moved to the end of the </body> tag and thus loses all styling of the surrounding .box element. Moving the styles out of the scope solves this and prevents duplicate styles.

kspearrin:
Ok. I'm not a big fan of the placement of the drag handles in the current screenshot since the minus icons are now misaligned with the "add new field" plus. I might play around with moving it so that the icons can still line up.

Are you able to add this to the browser and web projects as well?

Kovah:
Sure, no problem. Handles might be placed on the right side or wherever it makes sense.

> Are you able to add this to the browser and web projects as well?

Sure, will take a look at it as soon as possible.

Kovah:
Just played with the placement of the handle, looks best the far right side of the field box, in my opinion. Would use this for all implementations if you are okay with it.

Screenshot 2019-03-27 at 15 22 24

kspearrin:
yea i think i like that better

package-lock.json
```diff
@@ -1397,12 +1406,20 @@
       "dev": true
     },
     "async": {
-      "version": "2.6.1",
-      "resolved": "https://registry.npmjs.org/async/-/async-2.6.1.tgz",
-      "integrity": "sha512-fNEiL2+AZt6AlAw/29Cr0UDe4sRAHCpEHh54WMz+Bb7QfNcFw4h3loofyJpLeQs4Yx7yuqu/2dLgM5hKOs6HlQ==",
+      "version": "2.6.2",
```
kspearrin: Did you run some other npm updates here? Not sure why this package was bumped since it doesn't seem to be a dependency of cdk.

Kovah: NPM seems to update packages on its own despite running npm install only. Should I revert this?

kspearrin: Yes, please only make changes that are necessary for this PR.

PR #245, opened 2019-04-13
Prettier UI on macOS

Rughalt:
Main goal of this change is to remove ugly (at least in my opinion) title bar from application on macOS, which looks terrible when app is using dark and nord themes. This is a simple css change in desktop, with additional change in jslib (also a PR) to allow electron to hide titlebar on 'darwin' platforms.

Application is draggable by most of the empty spaces.

I tested it as well as I could, to make sure important elements do not cause window to be dragged. If there are any left, let me know. I hope i didn't break anything (I shouldn't really, this is just CSS).

Here are screenshots after changes:

bitwarden-macos-2
bitwarden-macos-1
bitwarden-macos-3
bitwarden-macos-4

### DIFF
--- a/src/scss/environment.scss
+++ b/src/scss/environment.scss
@@ -7,3 +7,33 @@
         border-top-color: #000000;
     }
 }
+
+html.os_macos {
+    body {
+        -webkit-app-region: drag
+    }
+
+    button, a, span, input, p, h1, h2, h3, h4, h5, h6, img, select, textarea, .box {
+        -webkit-app-region: no-drag;
+    }
+
+    #login-page, #register-page, #hint-page, #two-factor-page, #lock-page {
+        .content {
+            a.settings-icon {
+                position: absolute;
+                left: unset;
+                right: 10px;
+
+                span {
+                    margin-right: 8px;
+                    float: left;
+                }
+            }
+        }
+    }
+
+    #groupings {
+        padding-top: 24px;
+    }
+
+}


kspearrin:
Looks nice

kspearrin:
I made a few improvements. When logged in, only the empty region around the buttons is draggable. When not logged in, body is draggable.

I am seeing one bug though. When there are scrollbars, click scrollbar itself causes a drag instead of being able to move the bar with the mouse. Not sure how to exclude that.

Rughalt:
Thanks for accept end merge! I totally missed this, as I never click on scrollbars (they don't even appear most of the time when using Magic Mouse and Macbook touchpad). Sorry... 

I too could not find a way to make scrollbars clickable when in ` --webkit-app-region: drag`. I looked around how other electron software does the drag region, and it seems that the easiest way is to create drag region on the top of the app, covering search box (I'll still work), where we now that scrollbars won't appear, like Notable does is (red box shows dragable region):
Zrzut ekranu 2019-04-14 o 06 13 44

For testing, I did something like that

```
body::before {
    content:"";
    position: absolute;
    top: 0;
    left: 0;
    z-index: -1;
    width: 100%;
    height: 37px; /* this is the height of vault .mac-bar */
    background: transparent;
    -webkit-app-region: drag
}
```

This adds invisible drag bar to top of the app. If the rest of exclusions (`-webkit-app-region: no-drag;`) are application wide it does not break search box in vault top bar. It still seems to break scrollbar in vault details box, but only on top - this can be fixed by making width of the `before` element `calc(100% - 12px)`. Scrollbar still acts weird if clicked near the top of the app - it 'freezes' when dragged down about 36-37px - but it seems it is not because of that - it works the same without `before` element. This may be because electron with 'titleBarStyle: hiddenInset' does something that we can't control.

kspearrin:
Seems like a decent solution. Let's try it. Can you send another PR?

Rughalt:
Of course, I'll try to make one tomorrow when I have some free time.

Crocmagnon:
I do not see this in the mac app, hasn't this been released ?

PR #246, opened 2019-04-16
Additional CSS to make app look and behave better on macOS

Rughalt:
This is an extension to PR #245. As discussed, I added invisible bar at the top of the app on macOS platform, and in addition to that I moved some UI elements to move scrollbars out of draggable area:

- Moved add new element button near search bar
- Moved element actions buttons to top.

Everything is done via SCSS (I had to add helper function to be able to use different footer colors on macOS platform). I did some preliminary testing and everything should work OK. Buttons in draggable area work, and I do not see any anomalies. Elements seem to scale and resize well, scrollbars also appear work.

I hope this is acceptable (as this slightly changes UI between Windows/Linux/macOS versions).

Here are some screenshots:
Zrzut ekranu 2019-04-16 o 19 49 55
Zrzut ekranu 2019-04-16 o 19 49 24
Zrzut ekranu 2019-04-16 o 19 49 00


kspearrin:
I am not really a fan of having different layouts like this on macOS vs the others. The UI wasn't really designed for a top bar like that, so I'd like to think it over a bit more (for example, making the action bar blue with white buttons like that doesn't look that great IMO). In my follow up update, I added a "mac bar" to the area at the top of the groupings column (around the min/max/close bottons) that is draggable so I think that will suffice for now.

The only place that the body is still draggable is on the frontend layouts when locked/logged out. That one still suffers from the scrollbar dragging problem.

Rughalt:
Thanks for feedback, it’s very appreciated.

I understand your point about the UI differences - this is a major change and as I said done in a hacky way. I’ll work on it a little bit more even if it is for my own use (I was going to implement MacOS like light and dark theme and add auto toggle so app changes appearance when OS does).

As for a login/unlock screen I never knew they could be scrolled - maybe I didn’t make window small enough. For this I have no idea how to fix it except to create draggavle overlay that covers whole window with conditional css that would make it 22px or so smaller when meeting certain sizes (when scrollbar appear). I’ll try playing with it when I get home.

Rughalt:
I've checked out newest version from repo and couldn't force scrollbar to appear in frontend views (create account, login etc.). It seems minimum window size id big enough to hide scrollbars. When in settings view, scrollbar works OK (because of your fix to modal backdrop). Could you kindly provide repro steps, so I can try fix it?

kspearrin:
You should be able to just `Ctrl + =` to zoom in a lot, which will create vertical scrollbars.

Rughalt:
Oh, I never came to me you could zoom in an electron app like that... that *is* unfortunate, I'll see if anything can be done... In my initial test however, when I zoomed in with `Ctrl + =`, created scrollbars worked without problem, as if they were created "outside" the draggable region.
Zrzut ekranu 2019-04-22 o 19 58 10

Rughalt:
Thats great. 

I'll close this PR then and keep my macOS related changes in my fork as I wrote earlier (I'll probably build a binary out of it with name like macwarden if you don't have anything against it).

Crocmagnon:
I'd love to have these changes implemented in the official Bitwarden app for desktop. No offense @Rughalt, but I don't want/like to rely on someone's fork to manage my passwords.

Rughalt:
None taken! I totally understand it, I also would prefer these changes merged into an official client because managing fork - even if changes are only cosmetic, the source is available and you can see my commits, check them and build it yourself - introduce another level of uncertainty.

I have commits ready - not yet pushed - with code required to add vibrancy and cleaner UI (colors in header/moved footer, button sizes etc). I can push them if needed (I'll post screens when I get home)

Rughalt:
As promised, here are screenshots of what I currently have in my local branch. First column (categories) has transparent background to allow for vibrancy. I'm not pushing it yet as I would have to create PR for jslib to allow transparency on electron window creation.
Zrzut ekranu 2019-04-24 o 17 37 49
Zrzut ekranu 2019-04-24 o 17 38 40

Crocmagnon:
This looks fantastic and is really in line with macOS design standards 😮

Rughalt:
I created PR (PR 38 in wbitwarden/jslib, [here](https://github.com/bitwarden/jslib/pull/38) - this build failed becuse of function chenged in that PR) that is required by newest commit. Vibrancy is only set for `darwin` platform. 

One warning touch - when testing, if you launch app with dev tools open at any point, vibrancy **won't work**.

Crocmagnon:
Are there any news on this PR ?

lukescammell:
> As promised, here are screenshots of what I currently have in my local branch. First column (categories) has transparent background to allow for vibrancy. I'm not pushing it yet as I would have to create PR for jslib to allow transparency on electron window creation.
> Zrzut ekranu 2019-04-24 o 17 37 49
> Zrzut ekranu 2019-04-24 o 17 38 40

As a multi-platform user, I would support this as a massive improvement to usability in general and would love to see the button bar moved up top even on Windows.

I find that having the action buttons down at the bottom creates a lot of mouse movement from the top to the bottom of the app.

User Journey
1. type something in search (up high)
2. search gives you the result you want
3. select that result from near to the top of the results list (still up high)
4. copy username (up high)
5. copy password (up high)
6. edit (down low)
7. edit most common field - username, password or URI (up high)
8. save (down low)
9. edit again (down low)
10. share (down low)
11. pick organisation (up high)
12. pick collection/collections (could be anywhere, let's say mid)
13. scroll ALLLLL the way to the bottom (down low) - these save and cancel buttons should be on the same line as the "SHARE" title, just above the Organisation picker, and should not scroll/should be fixed position.

If we repeat that with everything in the top bar, like this:
1. type something in search (up high)
2. search gives you the result you want
3. select that result from near to the top of the results list (still up high)
4. copy username (up high)
5. copy password (up high)
6. edit (up high)
7. edit most common field - username, password or URI (up high)
8. save (up high)
9. edit again (up high
10. share (up high)
11. pick organisation (up high)
12. pick collection/collections (could be anywhere, let's say mid)
13. scroll ALLLLL the way to the bottom (up high) - these save and cancel buttons should be on the same line as the "SHARE" title, just above the Organisation picker, and should not scroll/should be fixed position.

Not only does your mouse/trackpad hand do less work going from high to low, but your eyes do less searching for the appropriate button, since everything is top loaded.

richiksc:
Could you use the system font (San Francisco) in the app to make it appear much more native? In CSS this would be `system-ui`.

eliykat:
Hi @Rughalt, thanks for your submission! If you're still interested in progressing this PR, please:
* sign the Contributor License Agreement (see the link provided by the CLAassistant bot above)
* reply here and tag me (@) so that I can mark it for further review

As this PR is more than 12 months old, I'll close it for now. However, this is just housekeeping. We encourage our community members to get involved and make contributions, so you can comment here even after it's closed and I'll reopen it if required.

Issue #244, opened 2019-04-12
Following URL breaks windows application


jpeizer:
After looking at it for a while I came to the conclusion that the following URL breaks the UI of the bitwarden windows application.

`https://www.paycomonline.net/v4/ee/web.php/app/login`

Setting this to the URl field will break the UI.

This does not appear to be restricted to a given Windows setup.

kspearrin:
This is a known problem at the moment. It has to do with the broken favicon that paycom uses, which crashes the renderer process in Electron. Should be fixed when we update to Electron 4.

nitrix:
As a temporary work-around, I think you could use the little gear icon and change the URL matching mode to prefix or regex until they update :)

kspearrin:
> As a temporary work-around, I think you could use the little gear icon and change the URL matching mode to prefix or regex until they update :)

Yes, that would stop the application from trying to render the website icon.

nitrix:
I deleted my comment to avoid speculating, but that's very useful @kspearrin. Much appreciated.

xtqqczze:
Perhaps we could temporarily disable website icons by default, as requested in #253? @kspearrin

xtqqczze:
Fixed in [v1.15](https://github.com/bitwarden/desktop/releases/tag/v1.15.2).

PR #253, opened 2019-05-06
Disable favicons by default

xtqqczze:
## Summary
* Disable website icons (by default) by setting disableFavicons to true in SettingsComponent.
* Stopgap solution until Electron is updated.
* See also issue #244.

## Context
Upon encountering a malformed favicon the renderer process crashes, leaving the application in an unusable state. Once the crash has occurred, the user cannot even disable website icons through settings. Since this issue results in the application being unavailable to the user, an urgent fix is required.

### DIFF
--- a/src/app/accounts/settings.component.ts
+++ b/src/app/accounts/settings.component.ts
@@ -31,7 +31,7 @@ import { Utils } from 'jslib/misc/utils';
 export class SettingsComponent implements OnInit {
     lockOption: number = null;
     pin: boolean = null;
-    disableFavicons: boolean = false;
+    disableFavicons: boolean = true;
     enableMinToTray: boolean = false;
     enableCloseToTray: boolean = false;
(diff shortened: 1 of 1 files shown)


Crocmagnon:
So what if a user enables this setting ? Wouldn't the application become unusable too ? If that's the case, then we shouldn't "solve" the issue this way.

xtqqczze:
> So what if a user enables this setting ? Wouldn't the application become unusable too ? If that's the case, then we shouldn't "solve" the issue this way.

With this change the application is at least usable with default settings. I have edited my summary to clarify that does PR does not in fix issue #244. @kspearrin has said this will likely be fixed when Electron is updated to version 4.

xtqqczze:
Issue #244 fixed in [v1.15](https://github.com/bitwarden/desktop/releases/tag/v1.15.2).

PR #257, opened 2019-05-31
add window option: always on top of other windows

Tobirexy:
Adds a checkbox option to the window menu which makes Bitwarden always stay on top of other windows. This is really useful when copying more than one thing to an other program because the window doesn't get minimized. This option is disabled by default.

Needed jslib changes: [https://github.com/bitwarden/jslib/pull/41](https://github.com/bitwarden/jslib/pull/41)

image

### DIFF
--- a/src/main/menu.main.ts
+++ b/src/main/menu.main.ts
@@ -450,8 +450,8 @@ export class MenuMain extends BaseMenu {
             {
                 type: 'checkbox',
                 label: this.main.i18nService.t('alwaysOnTop'),
-                checked: this.windowMain.win.isAlwaysOnTop() ? true : false,
-                click: () => this.main.messagingService.send('alwaysOnTop'),
+                checked: this.windowMain.win.isAlwaysOnTop(),
+                click: () => this.main.windowMain.toggleAlwaysOnTop(),
                 accelerator: 'CmdOrCtrl+Shift+T',
             },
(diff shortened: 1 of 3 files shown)


src/main/menu.main.ts
```diff
@@ -441,11 +441,20 @@ export class MenuMain extends BaseMenu {
...
+        (template[template.length - 2].submenu as MenuItemConstructorOptions[]).splice(1, 0,
+            {
+                label: this.main.i18nService.t('hideToTray'),
+                click: () => this.main.messagingService.send('hideToTray'),
+                accelerator: 'CmdOrCtrl+Shift+M',
+            },
+            {
+                type: 'checkbox',
+                label: this.main.i18nService.t('alwaysOnTop'),
+                checked: this.windowMain.win.isAlwaysOnTop() ? true : false,
```
kspearrin: I don't think ternary operator is necessary here

src/main/menu.main.ts
```diff
@@ -441,11 +441,20 @@ export class MenuMain extends BaseMenu {
...
+                click: () => this.main.messagingService.send('alwaysOnTop'),
```
kspearrin: Could it just call `this.main.windowMain.alwaysOnTop()` directly instead of sending a message?

PR #258, opened 2019-05-31
If an environment is configured, show the hostname in the lock screen

pachulo:
This PR needs this other to be merged first: https://github.com/bitwarden/jslib/pull/42

This tries to implement this feature request: https://community.bitwarden.com/t/show-the-server-to-which-you-are-logged-in/5472
By showing the hostname of the custom environment, if one is configured, in the lock screen.

### DIFF
--- a/src/locales/en_GB/messages.json
+++ b/src/locales/en_GB/messages.json
@@ -743,6 +743,19 @@
       }
     }
   },
+  "loggedInAsOn": {
+    "message": "Logged in as $EMAIL$ on $WEBVAULTHOSTNAME$.",
+    "placeholders": {
+      "email": {
+        "content": "$1",
+        "example": "name@example.com"
+      },
+      "webVaultHostname": {
+        "content": "$2",
+        "example": "bitwarden.example.com"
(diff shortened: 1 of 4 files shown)


kspearrin:
Could you share what the use case is here? Are you logged into multiple servers or something?

pachulo:
My use case here was having accounts in the public and the on-premises instances using the same email, so I did not know where I was logged-in when in locked state.
Besides that, I also think that this is useful for new users using an on-premises instance, so they know where they are logged.
And also: I wanted to try to contribute to Bitwarden and I thought that this was an easy thing to start with, so I did it.

src/locales/en_GB/messages.json
```diff
@@ -743,6 +743,19 @@
       }
     }
   },
+  "loggedInAsOn": {
```
kspearrin: do not modify this file, please. only modify the `en` file for new languages. these other get added by our translation tools

pachulo: OK, done!

PR #288, opened 2019-08-03
Add warning when navigating away from an edited entry

bigbwalter:
Added a warning for when the user attempts to navigate away from an entry that has been edited. This works on the add and edit screens. This is my first time contributing to an open source project, so please feel free to give me any and all feedback. Thanks!

### DIFF
--- a/src/app/vault/vault.component.ts
+++ b/src/app/vault/vault.component.ts
@@ -223,14 +223,19 @@ export class VaultComponent implements OnInit, OnDestroy {
         });
     }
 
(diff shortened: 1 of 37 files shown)


kspearrin:
You should only make changes to the `en` locale file. All others will be handled via the translation tools. Can you revert those, please?

bigbwalter:
I've removed the changes from all the locale files except for the en file. Let me know if there is anything else I can do!

kspearrin:
This was added in a recent update through other commits. Thanks.

PR #303, opened 2019-09-03
Only show carats for items with children

DarkArc:
This is a very small tweak, but it makes the app feel a lot higher quality.

ex:
image

### DIFF
--- a/src/app/vault/groupings.component.html
+++ b/src/app/vault/groupings.component.html
@@ -73,7 +73,7 @@
                     <ng-template #recursiveCollections let-collections>
                         <li *ngFor="let c of collections" [ngClass]="{active: c.node.id === selectedCollectionId}">
                             <a href="#" appStopClick appBlurClick (click)="selectCollection(c.node)">
-                                <i class="fa-li fa" title="{{'toggleCollapse' | i18n}}" aria-hidden="true"
+                                <i *ngIf="c.children.length" class="fa-li fa" title="{{'toggleCollapse' | i18n}}" aria-hidden="true"
                                     [ngClass]="{'fa-caret-right': isCollapsed(c.node), 'fa-caret-down': !isCollapsed(c.node)}"
                                     (click)="collapse(c.node)" appStopProp></i>
                                 {{c.node.name}}
--- a/src/app/vault/groupings.component.html
+++ b/src/app/vault/groupings.component.html
@@ -50,7 +50,7 @@
                     <li *ngFor="let f of folders"
                         [ngClass]="{active: selectedFolder && f.node.id === selectedFolderId}">
                         <a href="#" appStopClick appBlurClick (click)="selectFolder(f.node)">
-                            <i class="fa-li fa" title="{{'toggleCollapse' | i18n}}" aria-hidden="true"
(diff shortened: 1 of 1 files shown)


src/app/vault/groupings.component.html
```diff
@@ -73,7 +73,7 @@ <h2>{{'collections' | i18n}}</h2>
                     <ng-template #recursiveCollections let-collections>
                         <li *ngFor="let c of collections" [ngClass]="{active: c.node.id === selectedCollectionId}">
                             <a href="#" appStopClick appBlurClick (click)="selectCollection(c.node)">
-                                <i class="fa-li fa" title="{{'toggleCollapse' | i18n}}" aria-hidden="true"
+                                <i *ngIf="c.children.length" class="fa-li fa" title="{{'toggleCollapse' | i18n}}" aria-hidden="true"
```
DarkArc: I do not have, nor know how to create child collections, so this portion is untested with that case.

kspearrin:
Thanks for the PR. I realize this is subjective, but I just don't really agree with this approach. I feel the original way is more uniform.

DarkArc:
It's true that it's more uniform, however, it becomes hard to tell what folders do and do not have subfolders. It's also additionally not a pattern I've seen in any other user interface across my usage of Windows, Mac, and Linux for something that has no subelements.

It's open source -- thanks -- so I can do what I want; but, consider that this may be what the average person would expect, and the usability drawback to always presenting the arrow, even when it does nothing.

Truthfully, I'd prefer if the folders started closed, and then I could open the ones I cared about, of course, in that case, it's hard to know if there's a subfolder without either remembering, or clicking on the toggle.

One could add a setting for this, but, it doesn't seem like something worth adding a setting for.

wolfika:
Sorry for the +1 type of comment, but I too feel like it would be more natural/common if the carets would show up only for such items that do have children. One quick example I could find is from Windows file manager, where the behavior is the same the PR author described. See screenshot attached.

VirtualBox_Win10Eval_01_10_2019_17_19_33

I think from a UX perspective the current behavior feels weird, like there's a caret which hints that "hey, I have some subfolders", you click on it, it looks like it does something, but in fact, nothing happens.

stilettk:
@kspearrin Can you please clarify why you think that original way is more uniform? Given the arguments from the screenshot above. Thanks!

cscharf:
Hi @DarkArc , 

We've discussed this internally and would be willing to merge it in, thank you! There's just one thing, our web vault also uses the same logic for displaying the more uniform carats/arrows next to each folder; that repo would also need to be updated for consistency sake before merging this one in (so there's the same user experience between both clients).

Would you be willing to do that? see: https://github.com/bitwarden/web/blob/master/src/app/vault/groupings.component.html#L67

Thanks,
Chad

DarkArc:
@cscharf I think I already did? https://github.com/bitwarden/web/pull/410 :sweat_smile:

cscharf (approved):

cscharf:
So you did, sorry 😁

DarkArc:
All good, appreciate this change being made :)

PR #327, opened 2019-10-13
Allow special characters selection for pwd generation

GeorgySerga:
Hello! This is a small PR to allow selection of special characters during the password generation. A PR for jslib would be added separately (edit: https://github.com/bitwarden/jslib/pull/50).

The idea came after some resources did not allow using the password with special characters from bitwarden, though were requiring special characters in general (the set of allowed characters was limited). This led to changing pwd manually (to substitute some special characters) or just clicking "Regenerate" untill the right pwd would be created.

Below are the screenshot.
The first screen is with special characters off, untouched from the existing one:
Screen Shot 2019-10-13 at 9 46 41 AM
The second screen is  with an extra row of checkboxes, where characters for the password can be selected:
Screen Shot 2019-10-13 at 9 46 21 AM

I'm trying to contribute for the first time to bitwarden repos, would you please let me know if anything else is required, `CONTRIBUTING.md` seems a bit short on the guidelines.

### DIFF
--- a/src/app/vault/password-generator.component.html
+++ b/src/app/vault/password-generator.component.html
@@ -87,6 +87,13 @@
                                 <input id="special" type="checkbox" (change)="saveOptions()"
                                     [(ngModel)]="options.special">
                             </div>
+                            <div class="box-content-row checkbox-row" appBoxRow [hidden]="!options.special">
+                                <span *ngFor="let char of options.specialCharacters;">
+                                    <label for="specialCharactersSet">{{char}}</label>
+                                    <input id="specialCharactersSet" name="SpecialCharactersSet" type="checkbox" (change)="saveOptions()"
+                                    [(ngModel)]="options.specialCharactersValues[char]">
+                                </span>
+                            </div>
                         </div>
                     </div>
                     <div class="box" [hidden]="!showOptions">
--- a/src/scss/box.scss
+++ b/src/scss/box.scss
@@ -130,6 +130,24 @@
         }
     }
 
+    &.checkbox-row {
+        display: flex;
+        justify-content: flex-start;
+        flex-wrap: wrap;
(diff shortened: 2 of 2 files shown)


kspearrin:
Thanks for the PR.

This looks pretty good. I am wondering how we can make this checkbox interface work on other platforms, like mobile where we would want similar feature parity.

I wonder if just making the special character selection an input field where the user can type in whatever they like would be a better, more flexible approach. We get requests sometimes where users say "why isn't character __ available in the special character selection?". Allowing the user to input those vaults would seem to be more appropriate.

GeorgySerga:
Thank you for the reply @kspearrin !

I see your point, also thought about it, as functionality for password generation is shared through `jslib`. It might be a better idea to add "special characters" creation into the global preferences, as a global option, not into password generation window. Clicking (on desktop) or tapping (on mobile) is easier than entering text or scrolling selects full of options, which would not be logically sorted (what goes first '$' or '#'). The easiest "terminal" approach for adding new characters IMO would be to have a text input with all the possible special characters. But adding this to password generation modal could lead to mistakes and general overload.

This input would need to restrict addition of alphanumeric characters as well as duplicates.

To conclude, I think checkboxes are the easiest toggle for such preferences, adding extra options could be a separate PR though.

What do you think?

GeorgySerga:
@kspearrin Hello, just to update the status here, if you had a chance to take a look. I can add similar (the same) interface to `bitwarden/web` later on, if you like it. Thanks!

Hinton:
Hi,

We are currently in the process of archiving this repository and moving the code and future development to the [bitwarden/clients](https://github.com/bitwarden/clients) repository. This PR will be closed but feel free to re-open it again on [bitwarden/clients](https://github.com/bitwarden/clients).

PR #333, opened 2019-10-19
Enabled drag of username, password and totp in desktop app

vhschlenker:
In collaboration with the [pullrequest 51 at jslib](https://github.com/bitwarden/jslib/pull/51) this enables the possibilty to drag the username, password and totp from the desktop app into fields in other applications or anywhere else a drag can reach.
A possible use case for avoiding the clipboard here are clipboard synchronization apps, where a password should not appear.

### DIFF
--- a/src/app/vault/view.component.html
+++ b/src/app/vault/view.component.html
@@ -12,7 +12,7 @@
                 <!-- Login -->
                 <div *ngIf="cipher.login">
                     <div class="box-content-row box-content-row-flex" *ngIf="cipher.login.username">
-                        <div class="row-main">
+                        <div class="row-main" draggable="true" (dragstart)="setTextDataOnDrag($event, cipher.login.username)">
                             <span class="row-label">{{'username' | i18n}}</span>
                             {{cipher.login.username}}
                         </div>
@@ -24,7 +24,7 @@
                         </div>
                     </div>
                     <div class="box-content-row box-content-row-flex" *ngIf="cipher.login.password">
-                        <div class="row-main">
+                        <div class="row-main" draggable="true" (dragstart)="setTextDataOnDrag($event, cipher.login.password)">
                             <span class="row-label">{{'password' | i18n}}</span>
                             <div [hidden]="showPassword" class="monospaced">
                                 {{cipher.login.maskedPassword}}</div>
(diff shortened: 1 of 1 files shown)


kspearrin:
Trying to test this out, but it doesn't seem to be working. Does it work for you now after I have merged?

vhschlenker:
Yes, but I admit it is not perfect.
Especially the drop part can be a bit tricky, since not ever app supports that.
Electron apps often make problems, but native Inputs work pretty good. Eventhough the aiming with the cursor has to be done while the view is "blocked" with the preview of the moved field.
But as soon as the green plus appears at the cursor, the field can be dropped by releasing the mouse button.

I made a quick example video to demonstrate the basic idea: [YouTube](https://www.youtube.com/watch?v=rFFnar9bsnw&feature=youtu.be)

kspearrin:
I was able to perform the drag and drop using Notepad on Windows. However, it doesn't work in the VS Code editor window for whatever reason.

I see one major problem with this feature that we will have to figure out or we're going to have to revert it. I can no longer user select the data to copy/paste manually. Even though we provide a copy button for these fields already, users expect to also be able to highlight the text.

vhschlenker:
drag-handle
Ok, I understand.
Instead of making the whole field draggable, we could use the "arrows" icon from font-awesome and make only that draggable? Meaning you have to drag from this Icon, and the field is selectable as before.
The code would look something like that:
`<button class="row-btn" href="#" appStopClick draggable="true"
                                (dragstart)="setTextDataOnDrag($event, cipher.login.password)" role="button">
                                <i class="fa fa-arrows" aria-hidden="true"></i>
                            </button>`.
I'd gladly prepare another pull request and update the one in the browser repo if you are ok with the arrangement of the buttons and such.

kspearrin:
That's what I was thinking too, but I think we need something smaller. We can't afford the real estate to add another full icon to the right like that, especially on the browser extension. Maybe some kind of icon to the top left area of the row. Not sure...

vhschlenker:
As I played with some drag handles ([css-only](https://codepen.io/zachariab/pen/wkrbc) or [material](https://material.io/resources/icons/?search=drag&icon=drag_indicator&style=baseline)) neither of them really felt appropriate. But the css-only one sparked a pretty simple Idea:
Only row label itself could be dragable which is just indicated by giving it the CSS property of `cursor: move;`
This way, there is no need for any more screen real estate, and the functionality stays the same. The only downside, which was also present in the original solution I admit, is that it is not clear from the beginning that the item could be dragged without hovering over the field.

kspearrin:
Let's try adding the function to the label for now and see how that works. That would be acceptable.

Tobirexy:
Sorry to interrupt but would it be possible to make the copy buttons draggable as well?

Issue #259, opened 2019-06-02
No warning when navigating away from an edited entry


DewVinci:
After changing a password, if I navigate to a different item without clicking Save, the edit is lost. The app should **really** warn the user before losing this data.

I just changed a handful of passwords. On the 6th or 7th, I realized I hadn't been pressing the Save button. I had to go back to all those websites and use the "forgot password" flows.

PR #338, opened 2019-10-28
Warn of unsaved changes

afranco07:
Closes https://github.com/bitwarden/desktop/issues/259

Ran into this a couple times and wanted to fix it 😄 should work for both new and edited ones!

### DIFF
--- a/src/app/vault/vault.component.ts
+++ b/src/app/vault/vault.component.ts
@@ -226,13 +226,16 @@ export class VaultComponent implements OnInit, OnDestroy {
     async viewCipher(cipher: CipherView) {
         if (this.action === 'view' && this.cipherId === cipher.id) {
             return;
-        } else if (document.getElementsByClassName("ng-dirty").length > 0 && (this.action == 'add' || this.action === 'edit')) {
+        } else if (
+            (this.action == 'add' || this.action === 'edit')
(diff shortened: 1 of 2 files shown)


src/app/vault/vault.component.ts
```diff
@@ -223,9 +223,18 @@ export class VaultComponent implements OnInit, OnDestroy {
         });
     }
 
-    viewCipher(cipher: CipherView) {
+    async viewCipher(cipher: CipherView) {
         if (this.action === 'view' && this.cipherId === cipher.id) {
             return;
+        } else if (document.getElementsByClassName("ng-dirty").length > 0 && (this.action == 'add' || this.action === 'edit')) {
```
kspearrin: What if an input outside of the add/edit forms is dirty? Like the search box?

afranco07: @kspearrin good catch! I'm not too familiar with angular, is there a way to check the `AddEdit` components inputs/fields only?

kspearrin: You could just limit your `document` query to a section of the DOM tree only applicable to the add/edit pages.

afranco07: @kspearrin  Changed it to `document.getElementById('details').getElementsByClassName('ng-dirty')`

PR #231, opened 2019-03-20
Feature request/prototype: Hot-key window to get and save passwords in any situation

samdesota:
I switched from dashlane to bitwarden today because your open source tools allowed me to build out a feature I've wanted for a long time, a simple hot key window that allows me to add passwords and get passwords in any situation without relying on autocomplete extensions working properly. 

In this PR, by pressing `Ctrl/Cmd + Alt + Space` you get a simple window that allows just that:

Screenshot (7)

I'm happy with this for my personal use, but wanted to share in case you guys were interested in the idea for master, perhaps with a couple more features like enabling/disabling and customization of the shortcut.

PR #351, opened 2019-12-12
Add two globalShortcuts

marpo60:
Hi,

Thanks for this great project!

I was facing the same issue described here https://github.com/bitwarden/desktop/pull/231

I want to easily get to Search Vault and Password Generator functionalities.

I added two `globalShorctus` that would dispay the bitwarden main window or create it if its closed and navigate to that functionality 

```
Cmd+Shift+L -> Search Vault
Cmd+Shift+G -> Password Generator
```

Maybe this is useful on its own or maybe it's a good starting point.

Best

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,4 +1,4 @@
-import { app } from 'electron';
+import { app, globalShortcut } from 'electron';
 import * as path from 'path';
 
 import { I18nService } from './services/i18n.service';
@@ -42,6 +42,26 @@ export class Main {
         } else if (process.platform === 'linux' && process.env.SNAP_USER_DATA != null) {
             appDataPath = path.join(process.env.SNAP_USER_DATA, 'appdata');
         }
+
+        app.on('ready', () => {
+          globalShortcut.register('CommandOrControl+Shift+L', async () => {
+            if (this.windowMain.win === null) {
+              await this.windowMain.createWindow();
+            }
+
+            this.messagingService.send('focusSearch');
+            this.windowMain.win.show();
+          });
+
+          globalShortcut.register('CommandOrControl+Shift+G', async () => {
+            if (this.windowMain.win === null) {
+              await this.windowMain.createWindow();
+            }
+
+            this.messagingService.send('openPasswordGenerator');
+            this.windowMain.win.show();
+          });
+        });
 
         if (appDataPath != null) {
             app.setPath('userData', appDataPath);


kspearrin:
I am a little concerned about the shortcuts registered here. For example `Cmd+Shift+L` is already a shortcut for autofilling with the Bitwarden browser extension. Does that conflict with this?

marpo60:
Good point, I don't user the chrome extension.

I installed it to test if it was causing any conflicts and here are my findings
If you are on a chrome windows and press Cmd+Shift+L -> Autofill will be used
If you are outside a chrome window and press Cmd+Shift+L -> Global shorcut of the Desktop app will be used.

kspearrin:
This was just released and it looks like the `Cmd+Shift+L` shortcut is conflicting with autofill in Firefox.

fooness:
`Cmd+Shift+G` breaks lots of shortcuts on other applications. And it’s not even possible to disabled this via System Settings (on macOS) or in Bitwarden itself?

marpo60:
@kspearrin What do you think of adding an checkbox below Clear Clipboard in the Preferences of Bitwarden to enable the GlobalShorcuts?

image

fooness:
Global shortcuts should not be defined in an application itself.

On macOS it should go the native System Preferences -> Keyboard -> Shortcuts -> Services/Accessibility route in my opinion.

I’m sure there’s something similar to Linux and Windows.

kzshantonu:
Hi. After the latest update, I cannot use ctrl+shift+L in browsers to autofill anymore. Please add an option to disable the global hotkeys. Thanks

some1else:
Global hotkeys are a commonly used feature in other password management tools (Lastpass, One password) and are sorely missing from the configuration screen in Bitwarden. Lack of those shortcuts may well be a reason preventing somebody from switching to Bitwarden from similar apps.

Contrary to what @fooness had stated -- aside from not overriding application-level keyboard shortcuts -- using the services pane of global shortcuts is a tedious workaround involving Automator actions, not employed by many macOS users.

global-shortcuts

It's unfortunate to see commit [9201730](https://github.com/bitwarden/desktop/commit/9201730fe17098e85581d584249483e69dc82fc6) removing code from this PR. I sincerely hope that [PR #845](https://github.com/bitwarden/desktop/pull/845) reinstates this functionality in a way that facilitates usage both for people coming from other password management apps, as well as those used to invoking the app with the mouse or a launch query.

PR #388, opened 2020-02-22
sweetalert: move to sweetalert2.

MartB:
The styling got adjusted to stay as close as possible to the original sweetalert1 styles.
The only visible change is the button order, it is the same as in the web-vault now (OK - CANCEL instead of CANCEL - OK)

- Removed old postinstall gulp hack
- Added tsconfig type definition for sweetalert2 module typing.

## Screenshots
**Old**
old
---
**New**
new

Font-weight did not match the overall bitwarden ui styling in the old version, the new one is correct.

### DIFF
--- a/src/app/accounts/settings.component.ts
+++ b/src/app/accounts/settings.component.ts
@@ -5,7 +5,7 @@ import {
 
 import { ToasterService } from 'angular2-toaster';
 import { Angulartics2 } from 'angulartics2';
-import swal from 'sweetalert';
+import Swal from 'sweetalert2/src/sweetalert2.js';
 
 import { DeviceType } from 'jslib/enums/deviceType';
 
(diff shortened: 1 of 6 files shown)


src/app/accounts/settings.component.ts
```diff
@@ -5,7 +5,7 @@ import {
 
 import { ToasterService } from 'angular2-toaster';
 import { Angulartics2 } from 'angulartics2';
-import swal from 'sweetalert';
+import Swal from 'sweetalert2/src/sweetalert2.js';
```
kspearrin: Use braces. `import { Swal } from 'sweetalert2/src/sweetalert2.js';`

tsconfig.json
```diff
@@ -8,7 +8,7 @@
     "target": "ES2016",
     "allowJs": true,
     "sourceMap": true,
-    "types": [],
+    "types": ["sweetalert2"],
```
kspearrin: Still not sure I understand why this had to be added. Can't you just add the typings package for sweetalert2?

src/app/accounts/settings.component.ts
MartB: See https://github.com/bitwarden/web/pull/465#discussion_r382960307

tsconfig.json
MartB: Tbh I don't understand it either...
I just know that it fails to find the type as the imported file is not a "module" itself but a normal js file when I don't add it.

A separate typings package does not exist (if you meant sth like npm install @types/sweetalert2). It is part of the sweetalert2 package but does not get recognized when I don't add this "types" thingy. (I'm no node pro, so please tell me if you have any other ideas.)

Importing the sweetalert2 module (e.g `import Swal from 'sweetalert2'`) leads to it auto injecting the CSS file which breaks the style overrides.
One would need to add a bunch of strategic !important statements to the custom styling CSS which seemed excessive to me.

kspearrin:
@MartB

kspearrin:
@MartB 

We're working on cleaning up some governance items on old contributions from contributors. Would you be able to sign our CLA [here](https://cla-assistant.io/bitwarden/browser) so that we do not have to remove your past contributions from the codebase? Many thanks!

MartB:
@kspearrin Sure, no problem for these tiny changes!
"You have agreed to the CLA for multiple repositories or organizations" in "Version: 2020-05-09"

Regards,
Martin

Issue #74, opened 2018-03-22
Update download link for Mac


raufis27:
Please update Bitwarden-1.0.5.dmg to Bitwarden-1.1.0.dmg

kspearrin:
Whoops. Done. Thanks.

PR #390, opened 2020-02-23
Add minimize when copying to clipboard option

elias-pap:
Hello.

  With this feature, the user doesn't have to minimize the window after copying the credentials. It requires [PR #74](https://github.com/bitwarden/jslib/pull/74) to be merged.

This functionality is already supported on other password managers.


src/main/messaging.main.ts
```diff
@@ -23,6 +23,11 @@ export class MessagingMain {
                 this.main.menuMain.updateApplicationMenuState(message.isAuthenticated, message.isLocked);
                 this.updateTrayMenu(message.isAuthenticated, message.isLocked);
                 break;
+            case 'minimize':
+                if (this.main.windowMain.win != null) {
+                  this.main.windowMain.win.minimize();
```
kspearrin: format with 4 spaces indenting

src/app/vault/vault.component.ts
```diff
@@ -159,6 +163,7 @@ export class VaultComponent implements OnInit, OnDestroy {
                         if (this.cipherId != null && uCipher != null && uCipher.id === this.cipherId &&
                             uCipher.login != null && uCipher.login.username != null) {
                             this.copyValue(uCipher.login.username, 'username');
+                            this.minimizeIfNeeded();
```
kspearrin: So this feature should only work when using the keyboard shortcuts? What about when copying username/password to clipboard from other ways, like the view page or the context menu?

src/main/messaging.main.ts
elias-pap: Done.

src/app/vault/vault.component.ts
elias-pap: Implemented with 8849399. Part of this is implemented in [PR #74](https://github.com/bitwarden/jslib/pull/74).

src/app/vault/view.component.ts
```diff
@@ -48,4 +55,33 @@ export class ViewComponent extends BaseViewComponent implements OnChanges {
...
+            this.i18nService.t('valueCopied', this.i18nService.t(typeI18nKey)));
+        this.minimizeIfNeeded();
+
+        if (typeI18nKey === 'password') {
+            this.eventService.collect(EventType.Cipher_ClientToggledHiddenFieldVisible, this.cipherId);
+        } else if (typeI18nKey === 'securityCode') {
+            this.eventService.collect(EventType.Cipher_ClientCopiedCardCode, this.cipherId);
+        } else if (aType === 'H_Field') {
+            this.eventService.collect(EventType.Cipher_ClientCopiedHiddenField, this.cipherId);
+        }
```
kspearrin: instead of re-implementing all of this, you can just call `super.copy(...)` and then do you minimize stuff after that.

elias-pap: Implemented with 566e0c0. Also added minimizeOnCopyToClipboardKey constant to jslib of this repo, since it's desktop specific.

src/app/vault/view.component.ts
```diff
@@ -48,4 +55,17 @@ export class ViewComponent extends BaseViewComponent implements OnChanges {
         this.platformUtilsService.eventTrack('View Password History');
         this.onViewCipherPasswordHistory.emit(this.cipher);
     }
+
+    copy(value: string, typeI18nKey: string, aType: string) {
+        super.copy(value, typeI18nKey, aType);
+        this.minimizeIfNeeded();
+    }
+
+    public async minimizeIfNeeded(): Promise<void> {
```
kspearrin: don't need to specify `public`. Linter should be telling you this.

src/app/vault/view.component.ts
```diff
@@ -35,7 +41,8 @@ export class ViewComponent extends BaseViewComponent implements OnChanges {
         cryptoService: CryptoService, platformUtilsService: PlatformUtilsService,
         auditService: AuditService, broadcasterService: BroadcasterService,
         ngZone: NgZone, changeDetectorRef: ChangeDetectorRef,
-        userService: UserService, eventService: EventService) {
+        userService: UserService, eventService: EventService,
+        protected messagingService: MessagingService, protected storageService: StorageService) {
```
kspearrin: make these private.

src/locales/en_GB/messages.json
```diff
@@ -112,6 +112,12 @@
     "message": "Copy value",
     "description": "Copy value to clipboard"
   },
+  "minimizeOnCopyToClipboard": {
```
kspearrin: revert this file. Only need to make changes to the `en` locale.

src/app/vault/view.component.ts
```diff
@@ -48,4 +55,17 @@ export class ViewComponent extends BaseViewComponent implements OnChanges {
...
     }
+
+    copy(value: string, typeI18nKey: string, aType: string) {
+        super.copy(value, typeI18nKey, aType);
+        this.minimizeIfNeeded();
+    }
+
+    public async minimizeIfNeeded(): Promise<void> {
+        const shouldMinimize =
+            await this.storageService.get<boolean>(ElectronConstants.minimizeOnCopyToClipboardKey);
```
kspearrin: Format like this:

```suggestion
        const shouldMinimize = await this.storageService.get<boolean>(
            ElectronConstants.minimizeOnCopyToClipboardKey);
```

src/app/vault/vault.component.ts
```diff
@@ -660,6 +660,7 @@ export class VaultComponent implements OnInit, OnDestroy {
             this.platformUtilsService.copyToClipboard(value);
             this.toasterService.popAsync('info', null,
                 this.i18nService.t('valueCopied', this.i18nService.t(labelI18nKey)));
+            this.viewComponent.minimizeIfNeeded();
```
kspearrin: Does this function get called in other contexts outside of viewComponent? Shouldn't you only do this if `this.viewComponent != null && this.action === 'view'`?

src/app/accounts/settings.component.ts
```diff
@@ -120,6 +121,8 @@ export class SettingsComponent implements OnInit {
         this.locale = await this.storageService.get<string>(ConstantsService.localeKey);
         this.theme = await this.storageService.get<string>(ConstantsService.themeKey);
         this.clearClipboard = await this.storageService.get<number>(ConstantsService.clearClipboardKey);
+        this.minimizeOnCopyToClipboard =
+            await this.storageService.get<boolean>(ElectronConstants.minimizeOnCopyToClipboardKey);
```
kspearrin: Format like this:

```suggestion
        this.minimizeOnCopyToClipboard = await this.storageService.get<boolean>(
            ElectronConstants.minimizeOnCopyToClipboardKey);
```

kspearrin:
You also have a jslib conflict we need to figure out at some point.

src/app/vault/vault.component.ts
elias-pap: Good point. The null check is needed. However, the user might be editing a cipher and while editing, copy a username from another cipher. If the window wouldn't minimize, that's not consistent with the minimize option turned on, which is what the user expects. I would suggest to move the minimizeIfNeeded() method somewhere else.

elias-pap:
Hello,
  I've made the requested changes. As a reminder, [PR #74](https://github.com/bitwarden/jslib/pull/74) has to be merged for this to compile. The jslib conflict arises because jslib gets regularly pulled from its repo. When this happens, this branch stays behind.

src/app/vault/view.component.ts
elias-pap: Done.

src/app/vault/view.component.ts
elias-pap: Done.

src/locales/en_GB/messages.json
elias-pap: Okay.

src/app/vault/view.component.ts
elias-pap: Done.

src/app/accounts/settings.component.ts
elias-pap: Done.

src/main/window.main.ts
```diff
@@ -0,0 +1,21 @@
+import { ElectronConstants } from 'jslib/electron/electronConstants';
+
+import { MessagingService } from 'jslib/abstractions/messaging.service';
+import { StorageService } from 'jslib/abstractions/storage.service';
+
+import { WindowMain as BaseWindowMain } from 'jslib/electron/window.main';
+
+export class WindowMain extends BaseWindowMain {
```
kspearrin: We don't need a new window.main class like this. Your original solution was better. Just send a message from the angular components. We don't want angular components in the renderer process having a dependency on something related to the electron main process.

elias-pap: Sure.

PR #429, opened 2020-04-13
[Soft Delete] - Added trash to desktop app

cscharf:
## Objective
This adds the Trash bin and soft-delete/restore functionality to the Desktop client.

## Changes
- **ciphers.component.html**: Prevent adding a new cipher to the Trash directly
- **groupings.component.html**: Added Trash "folder" to top-level groupings navigation
- **vault.component.html**: Added vault orchestration event hooks/handlers for deleted and restored items
- **vault.component.ts**: Added deleted item view (trash) context tracking and state management. Also managed context menus for items in trash vs. not deleted
- **view.component.html**: Added delete and restore buttons

## Screenshots
Added the Delete button to the View, and the Trash "folder" to the groupings navigation along with contextual changes for deleted items.
image

### DIFF
--- a/src/app/vault/ciphers.component.html
+++ b/src/app/vault/ciphers.component.html
@@ -39,7 +39,7 @@
     </ng-container>
 </div>
 <div class="footer">
-    <button appBlurClick (click)="(deleted ? false : addCipher())" (contextmenu)="addCipherOptions()" 
+    <button appBlurClick (click)="addCipher()" (contextmenu)="addCipherOptions()" 
         class="block primary" appA11yTitle="{{'addItem' | i18n}}" [disabled]="deleted">
         <i class="fa fa-plus fa-lg" aria-hidden="true"></i>
     </button>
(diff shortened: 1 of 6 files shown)


src/app/vault/ciphers.component.html
```diff
@@ -39,8 +39,8 @@
     </ng-container>
 </div>
 <div class="footer">
-    <button appBlurClick (click)="addCipher()" (contextmenu)="addCipherOptions()" class="block primary"
-        appA11yTitle="{{'addItem' | i18n}}">
+    <button appBlurClick (click)="(deleted ? false : addCipher())" (contextmenu)="addCipherOptions()"
```
kspearrin, line 42: Is the `(click)` event still fired when disabled?

cscharf: No, I'll remove that... was my bad my [disabled] prop wasn't working, but I just wasn't patient enough for electron to rebuild.

cscharf: Corrected.

PR #457, opened 2020-05-27
updated icons

tgreer-bw:


kspearrin:
Are we missing the `resources/appx` updates?

tgreer-bw:
> Are we missing the `resources/appx` updates?

Yep, that was it. Added those in the latest commit.

PR #470, opened 2020-06-28
Biometric support

Hinton:
Initial draft on biometric support for desktop.

Currently supports Windows Hello. Still a few things left to sort out and need to verify everything compiles and runs on other platforms.


src/app/accounts/lock.component.html
```diff
@@ -36,5 +36,10 @@
                 {{'logOut' | i18n}}
             </button>
         </div>
+        <div class="button" *ngIf="supportsBiometric">
+            <a class="btn primary block" appBlurClick (click)="unlockBiometric()">
+                <i class="fa fa-unlock-alt" aria-hidden="true"></i> <b>{{'unlockWithWindowsHello' | i18n}}</b>
```
cscharf, line 41: Would we want this text and/or action to be OS/environment aware so that we weren't using a generic term like (unlock or supports Biometric) when we really mean, "Windows Hello" vs. "Touch ID" (macOS)?

Hinton: I was thinking on switching depending on platform. On windows it would say "Unlock with Windows Hello" while on Mac "Unlock with Touch ID". This initial draft is just a working version of Windows Hello.

src/locales/en/messages.json
```diff
@@ -1237,6 +1237,18 @@
...
+  },
+  "windowsHelloConsentMessage": {
+    "message": "Please authenticate yourself to unlock the Bitwarden Vault."
+  },
+  "unlockWithTouchId": {
+    "message": "Unlock with Touch ID"
+  },
+  "touchIdConsentMessage": {
+    "message": "Please authenticate yourself to unlock the Bitwarden Vault."
+  },
```
kspearrin, line 1251: Ony mobile we call this consent message the "direction" and just say "Use your fingerprint to verify." or "Use Face ID to verify.". Should we be consistent here? Something like "Use Windows Hello to verify." and "Use your fingerprint to verify". The consent message may not be displayed int eh context of unlocking the vault.

Hinton: I think we should include Bitwarden in the message since it's just a popup prompt. What about `Verify for Bitwarden.`?

kspearrin: That works

Issue #528, opened 2020-08-27
MacOS: Unable to drag window from rightmost pane.


jacobpledger:
## Describe the Bug
Using the desktop app on MacOS, if I go to move the window, I cannot initiate a drag from the top of the rightmost pane - where an actual entry would appear when selected. Additionally it is hard to grab the window from the middle pane because the search box is right at the top. This leaves the only safe space to drag the window from as the top of the rightmost pane, and this throws me off constantly. Previously, when the app had a title bar this was not an issue.

## Steps To Reproduce
On MacOS (probably any OS, but I haven't checked), launch the desktop app and login, then attempt to drag the window around the screen by grabbing the top of the window in each of its 3 vertical sections.

## Expected Result
You should be able to move the window around by grabbing anywhere along the top of the window, possibly other unpopulated areas too. For example, the lock screen of the desktop app can be moved by grabbing anywhere in the window.

## Actual Result
The window cannot be moved from the rightmost pane (entry view), and is difficult to grab from the top of the middle pane (list view of all logins)

## Environment

Mac OS Catalina
- Build Version 1.20.1

addisonbeck:
Currently, this is a consequence of the intended design. I agree that there should be more draggable space, but getting there is going to require a bit of a redesign. See the attached pull request for some more detail.

I'd suggest making this a feature request on our [community forum](https://community.bitwarden.com/) if you'd still like to see it implemented in the future. We use the forum to gauge overall interest in larger changes like this one. Thanks!

PR #561, opened 2020-09-30
added a draggable class for electron

addisonbeck:
Closes https://github.com/bitwarden/desktop/issues/528

### Issue 😧
Click and drag for MacOs is impossible from the vault screen except from a small area above the search box, even when there is ample other empty space around (like if no login is selected)

### Solution 💰
Add a css selector to signify the app as draggable

### Risk ⚠️
The [electron docs](https://www.electronjs.org/docs/api/frameless-window#draggable-region) on draggable-region mentions that if a you apply it to the entire document, as I have here, you need to specify certain fields like buttons as being not-draggable. I did not do anything for this, as all buttons, links, and text inputs in the app seemed to be working fine without it, but I would consider this a risk.

### DIFF
--- a/src/scss/base.scss
+++ b/src/scss/base.scss
@@ -16,7 +16,7 @@ html, body {
 body {
     color: $text-color;
     background-color: $background-color-alt2;
-
+    -webkit-app-region: drag;
     @include themify($themes) {
         color: themed('textColor');
         background-color: themed('backgroundColorAlt2');


src/scss/base.scss
```diff
@@ -16,7 +16,7 @@ html, body {
 body {
     color: $text-color;
     background-color: $background-color-alt2;
-
+    -webkit-app-region: drag;
```
kspearrin, line 19: Draggable styling for macos is defined in `environment.scss`. Changes should go there. We need these to be defined on elements explicitly to avoid weird behaviors around dragging.

kspearrin (changes requested):

src/scss/base.scss
addisonbeck: Thanks, I see that now. [This](https://github.com/bitwarden/desktop/blob/7499d7de2522fab2c05f02966efe1eb45c196f95/src/app/vault/vault.component.ts#L187) line of code is causing the body to lose the class defined for this when the vault loads, which is making the app lose all it's draggability. Any idea why this class the layout_frontend class is being removed here @kspearrin ?

kspearrin: That class is for "frontend" layouts (i.e. the layout on pages when you are not logged in). The draggable stuff was intentionally not made available on all areas of the window since there is a lot more stuff included on-screen whenever you log in. Usually, there is some kind of defined bar at the top or something that you use as a draggable area.

addisonbeck: For the mac desktop app, there isn't much. Draggable areas are limited to the spots outlined here. If that's the intended design we can leave this as is and I'll close the linked ticket. Let me know. 

Screen Shot 2020-10-01 at 12 34 04 PM

kspearrin: That is the intention today. I agree that it would be nice to have a larger area that is draggable, but I am not sure how to make that work with this design layout. We don't want to just all background areas draggable.

Issue #285, opened 2019-08-01
Close to tray does not work in macOS and linux


netvl:
The "Close to tray" feature implemented in https://github.com/bitwarden/desktop/pull/168 does not work on macOS or Linux.

In macOS, closing the main window does not exit the application (which is good), but it results in a necessity to re-type the master password, which is really unfortunate. It is interesting that clicking the tray icon does result in correct behavior (main window being hidden/showed again, without changing the locked state).

In Linux, closing the main window results in the entire application being closed. Naturally, after restart I have to enter the master password again. Ideally, in Linux with this option enabled should result into the main window being hidden, but the app itself should remain in memory and keep its locked/unlocked state.

LanceSandino:
+1 to this -- I mentioned this here: https://github.com/bitwarden/desktop/issues/283#issuecomment-517110848

but should have probably made a new issue for it :)

idiaz97:
It does not work on windows 10 either

Cris70:
It works here if I choose "Window" => "Hide to Tray" from the menu, but if I close the window using the X button the application is closed instead of hiding into the tray. I would expect it to hide instead, when in the options I have "Enable Tray Icon" selected.
I am on openSUSE Tumbleweed (KDE).

netvl:
Any chance to get this fixed? If not, are contributions accepted? I might have got enough resolution to try and fix it myself, although I never worked with electron before :)

sshermz:
I would like to move to BW on linux, but this is a major usability issue.

It should be trivial to fix as the functionality is already implemented on linux via Window|Hide to Tray.

'Close to Tray Icon' is present in the settings on Windows, but is missing in settings on Linux. So it looks like Linux was partially neglected when this feature was implemented  >_<

matte-Rhorn:
In the version for Windows there is a dedicated toggle button in the settings to manage the application behavior when the close and minimize buttons are clicked. That toggle button allows deciding if it should be minimized to the tray. Such option is missing in the version for Linux.

Would be great if such option 'Close to Tray Icon' button could be implemented also for the platforms other than Windows.

Cris70:
Any news on this?

silverscat-3:
I'm currently attempting to solve this problem; Close to tray works by commenting out the platform conditional branch in `tray.main.ts` (at least on Linux) but I can't found the code that hides that setting. It's too bad, I don't understand how the setting is hidden at the moment.

Cris70:
[...]

PR #569, opened 2020-10-13
Fix Close to tray does not work in macOS and linux #285

silverscat-3:
Tested in the following environment:

- **OS:** Arch Linux
- **WM:** Awesome

I haven't tested macOS because I don't own a Macintosh.

### DIFF
--- a/src/app/accounts/settings.component.html
+++ b/src/app/accounts/settings.component.html
@@ -47,7 +47,8 @@
                         <div class="form-group" *ngIf="supportsBiometric">
                             <div class="checkbox">
                                 <label for="biometric">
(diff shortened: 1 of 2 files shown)


cscharf (changes requested):
@silverscat-3 , I don't see a PR for `jslib` or a fork for it with the required changes necessary to facilitate these settings working. I've done a sanity check with your PR against macOS and it does not work, the original issue still persists as reported in the linked issue along with the undesired behavior. Please consider creating a PR for this from `jslib` with the suggested changes that are working in Linux for evaluation so that can be tested in tandem with this change.

silverscat-3:
@cscharf I understand. I have no plans to buy a Macintosh at the moment, so apparently, that's as far as I'm going to go.

I will consider creating a PR on `jslib`.

PR #577, opened 2020-10-29
Handle main window hidden event

cscharf:
## Overview
When minimizing, occluding, hiding, closing to tray, etc. any hidden fields on the screen such as passwords, card codes, hidden custom fields, etc. should be re-hidden if they were being viewed/visible before the app/screen was hidden.

## Changes
* **main.ts** - Handle the `BrowserWindow`'s `hide` and `minimize` events which across all platforms generally capture at the least when a window is minimized or closed to the tray; on macOS `hide` is also invoked when the window is occluded. These events can and will fire multiple times for certain actions like minimizing the window on some platforms so care was taken that no heavy processing or destructive actions were taken
* **[vault|account].component.ts** - Handle the new message service broadcast with command = `"windowHidden"`
   * `onWindowHidden()` event handler method to set all `showPassword/Value/Code` to `false`
   * Added the `BroadcastService` and `NgZone` (where necessary) to handle the new event broadcast and ensure proper zone context
   * Added const, `BroadcasterSubscriptionId` to mimic the pattern implemented in other components which use the broadcast service
   * Where missing/not implemented, added `NgInit` and `ngOnInit` methods; otherwise overrode those calling `super` first, then adding necessary broadcaster subscription

### DIFF
--- a/src/app/accounts/lock.component.ts
+++ b/src/app/accounts/lock.component.ts
@@ -1,4 +1,8 @@
-import { Component } from '@angular/core';
+import {
+    Component,
+    OnDestroy,
+    NgZone,
+} from '@angular/core';
 import {
     ActivatedRoute,
     Router,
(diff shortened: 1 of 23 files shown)


src/app/accounts/lock.component.ts
```diff
@@ -39,5 +47,19 @@ export class LockComponent extends BaseLockComponent {
                 setTimeout(() => this.unlockBiometric(), 1000);
             }
         });
+        this.broadcasterService.subscribe(BroadcasterSubscriptionId, async (message: any) => {
```
kspearrin, line 50: You need to unsubscribe from the broadcaster service onDestory. See examples in other components.

kspearrin (changes requested):

src/app/accounts/lock.component.ts
cscharf: 🤦🏻 , I had missed that before, got that implemented now.

kspearrin (approved):

kspearrin (approved):

PR #579, opened 2020-11-01
[GDPR] Adjusted TOS/Privacy acceptance

vincentsalucci:
## Code Changes
- **register.component.html**: Added new checkbox and label elements. Styled similarly to other footer text within this component.
- **messages.json**: Added label message and error message for policy acceptance.
- **jslib**: Update from d84d6da to 5e50aa1

## Screenshots
desktop-0-base
desktop-1-error

### DIFF
--- a/src/app/accounts/register.component.ts
+++ b/src/app/accounts/register.component.ts
@@ -16,11 +16,25 @@ import { RegisterComponent as BaseRegisterComponent } from 'jslib/angular/compon
     templateUrl: 'register.component.html',
 })
 export class RegisterComponent extends BaseRegisterComponent {
+    showTerms = true;
+    acceptPolicies: boolean = false;
+
     constructor(authService: AuthService, router: Router,
         i18nService: I18nService, cryptoService: CryptoService,
         apiService: ApiService, stateService: StateService,
         platformUtilsService: PlatformUtilsService, passwordGenerationService: PasswordGenerationService) {
         super(authService, router, i18nService, cryptoService, apiService, stateService, platformUtilsService,
             passwordGenerationService);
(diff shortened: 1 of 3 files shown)


src/app/accounts/register.component.ts
```diff
@@ -16,11 +16,25 @@ import { RegisterComponent as BaseRegisterComponent } from 'jslib/angular/compon
     templateUrl: 'register.component.html',
 })
 export class RegisterComponent extends BaseRegisterComponent {
+    showTerms = true;
+    acceptPolicies: boolean = false;
+
```
kspearrin, line 21: Would it not have been better to put this all in jslib base component?

vincentsalucci: It was a passing thought for me and it is possible. The reason I didn't do it, is because it does not apply to all platforms using jslib. 

Can easily create a jslib PR with these tracking variables and the setting of `showTerms` in the class constructor. Would leave the submit override in each necessary platform, however.

Thoughts?

kspearrin: Which platforms does this not apply to? I can't think of any now.

vincentsalucci: CLI and Directory Connector

kspearrin: Neither of those uses this component (CLI isn't angular and neither offers registration). I think we can move all logic into jslib base registration component.

vincentsalucci: On it!

kspearrin (dismissed):

kspearrin (approved):

PR #614, opened 2020-12-14
Fix autostart in snap

Hinton:
This resolves the autostart not working correctly in snap packages on linux.

Also ensures that "Enable tray icon" is enabled when using other tray related settings on linux.
image

### DIFF
--- a/src/app/accounts/settings.component.ts
+++ b/src/app/accounts/settings.component.ts
@@ -53,6 +53,7 @@ export class SettingsComponent implements OnInit {
     alwaysShowDock: boolean;
     showAlwaysShowDock: boolean = false;
     openAtLogin: boolean;
+    requireEnableTray: boolean = false;
 
     enableTrayText: string;
     enableTrayDescText: string;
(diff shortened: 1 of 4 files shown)


src/app/accounts/settings.component.ts
```diff
@@ -269,17 +273,40 @@ export class SettingsComponent implements OnInit {
...
+        }
+
         await this.storageService.save(ElectronConstants.enableTrayKey, this.enableTray);
         this.callAnalytics('Tray', this.enableTray);
         this.messagingService.send(this.enableTray ? 'showTray' : 'removeTray');
     }
 
     async saveStartToTray() {
+        if (this.requireEnableTray) {
+            this.enableTray = true;
```
cscharf, line 307: Since this method isn't also persisting `this.enableTray` to the storage service, would we lose that setting if the user simply exited the app or negated that screen without saving tray settings?

Hinton: Whops, correct, we should save it.

cscharf (approved):

PR #615, opened 2020-12-16
Native Messaging - Support multiple extensions concurrently

Hinton:
The old implementation assumed that only a single browser extension was running, which resulted in the `sharedSecret` being constantly invalidated. This change assumes the browser sends a unique `appId` which is used to decide which shared secret to use.

Also display an error message informing the user to turn on desktop biometrics before browser biometrics can be enabled.
Windows biometric prompt now grabs focus to ensure it appears above browser windows. 

Depends on https://github.com/bitwarden/jslib/pull/234


src/locales/en/messages.json
```diff
@@ -1453,9 +1453,15 @@
   "verifyBrowserTitle": {
     "message": "Verify browser connection"
   },
-  "verifyBrowserDescription": {
+  "verifyBrowserDesc": {
```
cscharf, line 1456: Why the change to an abbreviation? Does it need to change?

Hinton: We use the abbreviation for all the other dialogs, so seemed best to be consistent.

cscharf (commented):
Looks good so far, waiting for final changes.

src/main/nativeMessaging.main.ts
```diff
@@ -19,15 +20,16 @@ export class NativeMessagingMain {
 
         ipc.serve(() => {
             ipc.server.on('message', (data: any, socket: any) => {
-                // This is a ugly hack until electron is updated 7.0.0 which supports ipcMain.invoke
```
cscharf, line 22: Does it make sense to remove this comment now?

Hinton: Yes, the implementation was changed slightly in order to support sending multiple messages to the browser instead of only responses. Which is needed to let the browser know we want to validate the fingerprint.

cscharf (dismissed):

cscharf:
@Hinton , can you update jslib for this so when merged will successfully build? The jslib change has been approved and merged.

Hinton:
@cscharf Done, thanks.

src/locales/en/messages.json
```diff
@@ -1453,15 +1453,33 @@
...
     "message": "Please ensure the shown fingerprint is identical to the fingerprint showed in the browser extension."
   },
+  "biometricsNotEnabledTitle": {
+    "message": "Biometrics not enabled"
+  },
+  "biometricsNotEnabledDesc": {
+    "message": "Browser biometric requires desktop biometric to be enabled in the settings first."
```
cscharf, line 1481: Should read: `"Browser biometrics requires desktop biometrics to be enabled in the settings first."` (biometrics should be plural)

cscharf (commented):
one quick change needed, either missed it from before or it reverted.

cscharf (approved):

PR #649, opened 2021-01-19
Tray Icon Mini UI

sarthurdev:
Thread: https://community.bitwarden.com/t/tray-icon-mini-ui/16792

mac-item
mac-locked


flying-sausages:
Any chance you could make the icon monochrome and fit the rest of the menu bar item appearances while you're at it? Not sure if it's scope creep or not.

sarthurdev:
@flying-sausages I think that probably would fall out of scope for this PR. I'll take a look though and open another PR if I can do something about it.

luximus-hunter:
This may be a stupid question, but does it work with Windows as well?

sarthurdev:
@luximus-hunter It does also work on Windows.

win-item
win-locked

flying-sausages:
I would generally argue that the "format" of that view should probably look more like the browser extension, there's a lot of valuable space which is being used for emptyness.

The usecase probably is more "Find a password real fast, copy it and get out" rather than "I need the full app from the taskbar".

Just some food for thought.

sarthurdev:
I agree, but the main window UI breaks when the dimensions were made smaller shown above. I don't have the time to work on this further to allow for a smaller profile on this tray icon.

I thought this would be a neat feature for Bitwarden and bring it further in line with competing products, but I haven't heard anything from anyone on the team.

luximus-hunter:
It might be best to make a separate UI for the task tray icon. Probably a clone or something similar to the browser extention like flying-sausages said.

PR #678, opened 2021-01-25
Add entitlements to mac, disable browser integration on Windows Store

Hinton:
Adds the missing entitlements to the MAS version, which should resolve browser integration not working on Mac. (This is untested since I haven't found any way test it locally)

Also disables browser integration in the Windows Store version for now. There are potential workarounds we could investigate but they are complex and some would require special approval to publish in the Microsoft Store.

image

### DIFF
--- a/resources/entitlements.mas.plist
+++ b/resources/entitlements.mas.plist
@@ -8,5 +8,10 @@
 	<true/>
 	<key>com.apple.security.files.user-selected.read-write</key>
 	<true/>
+	<key>com.apple.security.temporary-exception.files.home-relative-path.read-write</key>
+	<array>
+		<string>/Library/Application Support/Mozilla/NativeMessagingHosts/</string>
+		<string>/Library/Application Support/Google/Chrome/NativeMessagingHosts/</string>
+	</array>
 </dict>
 </plist>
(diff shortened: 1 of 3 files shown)


cscharf (approved):

resources/entitlements.mas.plist
```diff
@@ -8,5 +8,10 @@
 	<true/>
 	<key>com.apple.security.files.user-selected.read-write</key>
 	<true/>
+	<key>com.apple.security.temporary-exception.files.home-relative-path.read-write</key>
+	<array>
+		<string>/Library/Application Support/Mozilla/NativeMessagingHosts/</string>
+		<string>/Library/Application Support/Google/Chrome/NativeMessagingHosts/</string>
```
kspearrin, line 14: @Hinton Do we need paths for other browsers here too? like edge chromium and opera?

Hinton: They should use the same as chrome in my experience. Not 100% sure on Edge though.

Hinton: Seems like edge on mac uses a different directory, I'll make a follow up PR for this.

PR #708, opened 2021-02-03
Fix electron 11 support

proletarius101:
Bitwarden has updates its electron dependency to 11 the other day. However, 
- keytar is an old version that doesn't support electron 11
- desktop-idle doesn't support electron 11 by far

So it leads to the problem https://github.com/flathub/com.bitwarden.desktop/pull/52#issuecomment-772369527

This PR updates keytar to the latest release and replaces desktop-idle with [powerMonitor.getSystemIdleTime()](https://www.electronjs.org/docs/api/power-monitor#powermonitorgetsystemidletime)

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,7 +177,6 @@
       "entitlementsInherit": "resources/entitlements.mas.inherit.plist",
       "hardenedRuntime": false,
       "asarUnpack": [
-        "node_modules/desktop-idle",
         "node_modules/keytar"
       ]
     },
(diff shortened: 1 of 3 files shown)


proletarius101:
BTW, there are so many legacy dependencies and npm reports `found 533 vulnerabilities (523 low, 2 moderate, 8 high)`

package-lock.json
```diff
@@ -3626,11 +3626,6 @@
         "minimalistic-assert": "^1.0.0"
```
kspearrin, line 3626: please revert your change to package-lock.json. We will run an update to this ourselves once merged.

kspearrin (changes requested):

package-lock.json
proletarius101: Done

kspearrin (approved):

PR #710, opened 2021-02-03
Send Layout & List

addisonbeck:
## Code Changes
- **jslib**: updated to latest
- **app-routing.module.ts**: added Send routes 
- **app.module.ts**:  added new Send components and the DatePipe
- **send/add-edit.component.***: created a blank component
- **send/send.component.***: created a nav bar and data list similar to the typical vault. Currently supports getting sends from the server, syncing, searching, and filtering.
- **vault/groupings.component.html**: added a nav bar.
- **vault/vault.component.html**:  Added classes to several large components for scss changes below.
- **en/messages.json**:  all the strings!
- **scss/buttons.scss**: created a new button style for navigating to and from Send
- **scss/*.scss**: changed several id selectors to class selectors to better convey that they are reused (in Send)
- **layout/nav.component.***: created a nav bar for toggling between Vault/Send and any future stuff.

## Testing Considerations
- A link was added to the Vault screen for navigating to Send and vice verse
- The Send view currently supports displaying existing sends in a list, searching, filtering, and syncing. Any other functionality should not be considered testable yet.
- The UI of the Vault screen was slightly modified. Elements in the left nav were moved around and some selector changes were made in styling. 

## Screenshots
Screen Shot 2021-02-05 at 12 48 02 PM
Screen Shot 2021-02-05 at 12 48 12 PM


kspearrin:
Not a bad idea on navigation between Vault and Send. To make that a bit more future-proof, would we move to a tab layout at the bottom of that column there that shows both vault and Send icons. I think this would jive a bit better with the style we have in other apps.

src/locales/en/messages.json
```diff
@@ -1497,5 +1497,21 @@
...
+  },
+  "sendTypeText": {
+    "message": "Text"
+  },
+  "searchSends": {
+    "message": "Search Sends",
+    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
+  },
+  "sendInformation": {
+    "message": "Send Information"
```
cscharf, line 1515: man this is messing with me, is this labeling a "View Send"? If not, maybe "Send: Information" or "Send, Information", not sure. It's probably fine, but no matter what you need to copy the `"description"` above to anywhere we use the term "Send"

cscharf (commented):
I also like the navigation concept, and agree perhaps a tabs approach would be great!

src/locales/en/messages.json
addisonbeck: Yes, that's in there right now for "View Send". In the Vault that page is labeled with "Item Information", so this is derivative of that. Luckily that string is sort of ahead of this PR, as "Viewing" a send is really just a screen that says "Send Information" on it right now.

addisonbeck:
@kspearrin @cscharf thoughts on this tab navigation design? Opinions on labels, or vertical centering, etc.? 

Screen Shot 2021-02-04 at 1 38 09 PM

cscharf:
> thoughts on this tab navigation design? Opinions on labels, or vertical centering, etc.?

I think making them "flex" span across, w/ centered icons would be awesome; any new tabs we add just shrinks all of them down evenly, so here they'd flex at 50%, add a 3rd then flex to 33%, etc. 

We would probably want labels under each icon since we have them in other clients (see Browser for instance). We just want to be cognizant of vertical space since not everyone maximizes the window or has it that large, in its default state with a good number of folders, etc. it should look natural, etc.

kspearrin:
@addisonbeck Let's make it a bottom pinned bar similar to the bar next to it that the add button is in. Then tabs inside of that.

src/locales/en/messages.json
addisonbeck: I added the "do not translate" description, but left the text as is for now since it isn't part of a testable feature yet.

src/app/app-routing.module.ts
```diff
@@ -15,6 +15,7 @@ import { SsoComponent } from './accounts/sso.component';
 import { TwoFactorComponent } from './accounts/two-factor.component';
 
 import { VaultComponent } from './vault/vault.component';
+import { SendComponent } from './send/send.component';
```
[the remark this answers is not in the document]
addisonbeck: Reordered, but I also just separated `send` and `vault` into separate groups. All other top-level folder have their own group here.

cscharf (approved):
Looks good to me, thanks Addison!

PR #720, opened 2021-02-09
View Sends

addisonbeck:
## Code Changes
- **jslib**: updated to latest
- **add-edit.component.***: set up a layout for displaying all Send fields deriving style from Send on `web` and the `desktop` Vault. Also added new service for updated `jslib`.
- **send.component.***: updated `ngOnInit` and dependencies per `jslib` updates. Changed the `editSend` function to just accept an id since that is all it needs.
- **messages.json**: copy/pasted a bunch of strings from `web`
- **box.scss**: added a `background-color` change to all `.box-content-row` children that matches the `background-color` applied to itself. I noticed in `<textarea>`s the `<textarea>` itself was never changing `background-color` with `focus/hover/active`.
- **misc.scss**: added s `.subtext` class to try and replicate the hint test used under some fields in `web` Send
- **variables.scss**: added a variable to facilitate `.subtext`

## Known Issues
- The date fields use `type="datetime-local"`. This is what the web Vault is using, but it is not themeable. The icon is always black and the calendar has a generic white/blue theme. This doesn't look very good on non-light themes.

## Testing Considerations
- This PR gives the ability to view existing sends created in another client. There is no save button involved here yet.

## Screenshots
Screen Shot 2021-02-09 at 3 36 38 PM
Screen Shot 2021-02-09 at 3 36 48 PM

### DIFF
--- a/src/locales/en/messages.json
+++ b/src/locales/en/messages.json
@@ -1515,11 +1515,58 @@
     "message": "Search Sends",
     "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
   },
-  "sendInformation": {
-    "message": "Send Information",
+  "editSend": {
+    "message": "Edit Send",
     "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
   },
   "myVault": {
     "message": "My Vault"
+  },
+  "text": {
+    "message": "Text"
+  },
+  "deletionDate": {
+    "message": "Deletion Date"
+  },
+  "deletionDateDesc": {
+    "message": "The Send will be permanently deleted on the specified date and time.",
+    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
+  },
+  "expirationDate": {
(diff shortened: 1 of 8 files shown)


src/locales/en/messages.json
```diff
@@ -1515,11 +1515,58 @@
...
+  },
+  "expirationDateDesc": {
+    "message": "If set, access to this Send will expire on the specified date and time.",
+    "description": "'Send' is a noun and the name of a feature called 'Bitwarden Send'. It should not be translated."
+  },
+  "maxAccessCount": {
+    "message": "Maximum Access Count"
+  },
+  "maxAccessCountDesc": {
+    "message": "If set, users will no longer be able to access this send once the maximum access count is reached.",
```
cscharf, line 1546: I know you copied these, but maybe we missed this before and it needs to be corrected. `Send` should be capitalized everywhere consistently, here's it's lowercase and should be changed.

cscharf (commented):
Just 1 small change I saw

cscharf (approved):

PR #722, opened 2021-02-16
Finished CRUD operations for Send

addisonbeck:
### jslib
- update to latest

### add-edit.component
- Cleared `password` on refresh since we don't ever show that value after creation
- Removed the `<ng-container>` wrapping the form and moved its `*ngIf` to the form's `.inner-content` so the page layout doesn't flicker in and out of existence when changing operations. 
- Added support for creating Sends within the form (create specific inputs, etc.).
- Added support for a cancel button that clears the form & sets its action to `None`. This works the same way the Vault cancel button does.

**NOTE:** The File/Text switcher is a radio button. Mobile & (I think) browser are using a bootstrap toggle switcher. Because `web` currently is not using a toggle, and bootstrap isn't used on desktop, desktop is currently using radio buttons.

### send.component
- Added `onCancel`, `onSaveSend` and `onDeletedSend` output handlers to refresh the screen with relevant data from those actions.
- Added sync support. This was in jslib, but was removed for client specific logic elsewhere. The implementation here is exactly as it was previously in jslib.
- Returned from `selectSend()` if the clicked Send is already selected to prevent the screen from trying to refresh unnecessarily. 

### messages.json
- Added more strings, all of which existed already in the web client.

### box.scss
- Added an exception to the default `.box label` styling because it does not work with radio buttons. This should probably get a genuine refactor at some point, but the changes required seemed out of scope for the Send implementation.
- Added `radio` type exceptions to any rules that already had `checkbox` type exceptions
- Added radio button stylings in the form of `.box-content-row-radio`, similar to how other input type rows get styled.

### Completed Functionality:
- Navigating between My Vault and Send
- Viewing the list of Sends
- Filtering Sends by type
- Searching for Sends
- Viewing a Send's details
- Creating a new Send
- Editing an existing Send
- Deleting a Send
- Live sync
- Cancelling an add/edit action

### Features Not Ready For Testing
- Using menu bar functions (password generator, password history etc.) from the Send tab.
- Right-click menu for Sends similar to what is present in Ciphers
- Status badges in the Send list (disabled, password protected, etc.)
- A share button
- Recent design requirements changes (hiding optional fields by default, copy link on save option, show password).
- Blocking File sends behind a premium subscription
- Disabled Send policy implementation

## Demo
https://user-images.githubusercontent.com/15897251/108104274-187a4a80-7059-11eb-8596-8faac64383a3.mov

### DIFF
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -67,7 +67,7 @@
                         </select>
                         <div class="subtext">{{'deletionDateDesc' | i18n}}</div>
                     </div>
-                    <div class="box-content-row" *ngIf="deletionDateSelect === 0">
+                    <div class="box-content-row" *ngIf="deletionDateSelect === 0 && !editMode">
                         <input id="deletionDateCustom" type="datetime-local" name="deletionDate" 
                             [(ngModel)]="deletionDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
                     </div>
@@ -80,7 +80,7 @@
                         </select>
                         <div class="subtext">{{'expirationDateDesc' | i18n}}</div>
                     </div>
-                    <div class="box-content-row" *ngIf="expirationDateSelect === 0">
+                    <div class="box-content-row" *ngIf="expirationDateSelect === 0 && !editMode">
                         <input id="expirationDateCustom" type="datetime-local" name="expirationDate" 
                             [(ngModel)]="expirationDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
                     </div>
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -1,106 +1,159 @@
-<ng-container *ngIf="send">
-    <form (ngSubmit)="submit()" [appApiAction]="formPromise">
-        <div class="content" *ngIf="send">
-            <div class="inner-content">
-                <div class="box">
-                    <div class="box-header">
-                        {{'editSend' | i18n}}
+<form (ngSubmit)="submit()" [appApiAction]="formPromise">
+    <div class="content">
+        <div class="inner-content" *ngIf="send">
+            <div class="box">
+                <div class="box-header">
+                    {{title}}
+                </div>
+                <div class="box-content">
+                    <div class="box-content-row" appBoxRow>
+                        <label for="name">{{'name' | i18n}}</label>
+                        <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus>
                     </div>
-                    <div class="box-content">
-                        <div class="box-content-row" appBoxRow>
-                            <label for="name">{{'name' | i18n}}</label>
-                            <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus>
+                </div>
+            </div>
+            <div class="box">
+                <div class="box-content">
+                    <ng-container *ngIf="!editMode">
+                        <div class="box-content-row box-content-row-radio">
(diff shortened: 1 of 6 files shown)


src/app/send/add-edit.component.html
```diff
@@ -1,106 +1,159 @@
...
+                                <label class="label-hack" for="type_{{o.value}}">
+                                    {{o.name}}
+                                </label>
+                            </div>
                         </div>
-                        <div class="box-content-row" appBoxRow *ngIf="send.type === sendType.File">
+                        <div *ngIf="send.type === sendType.File" class="box-content-row" appBowRow>
+                            <label for="file">{{'file' | i18n}}</label>
+                            <input type="file" id="file" class="form-control-file" name="file" required
+                                [disabled]="disableSend">
```
MGibson1, line 32: It looks like this is the only place the disable send policy is being honored (and working). It's OK to ignore, I'll get to it once the feature is merged in, but it might make sense to remove these references until they're all implemented

src/app/send/add-edit.component.html
```diff
@@ -1,106 +1,159 @@
...
-                            <div class="subtext">{{'deletionDateDesc' | i18n}}</div>
-                        </div>
-                        <div class="box-content-row" appBoxRow>
-                            <label for="expirationDate">{{'expirationDate' | i18n}}</label>
-                            <input id="expirationDate" type="datetime-local" name="expirationDate" [(ngModel)]="expirationDate">
-                            <div class="subtext">{{'expirationDateDesc' | i18n}}</div>
-                        </div>
+                    <div class="box-content-row" *ngIf="deletionDateSelect === 0 && !editMode">
+                        <input id="deletionDateCustom" type="datetime-local" name="deletionDate" 
+                            [(ngModel)]="deletionDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
```
MGibson1, line 72: This readonly tag doesn't appear to do anything in my attempts

src/app/send/add-edit.component.html
```diff
@@ -1,106 +1,159 @@
...
-                <div class="box">
-                    <div class="box-content">
-                        <div class="box-content-row" appBoxRow>
-                            <label for="password">{{(hasPassword ? 'newPassword' : 'password') | i18n}}</label>
-                            <input id="password" type="password" name="password" [(ngModel)]="password">
-                            <div class="subtext">{{'sendPasswordDesc' | i18n}}</div>
-                        </div>
+                    <div class="box-content-row" *ngIf="expirationDateSelect === 0 && !editMode">
+                        <input id="expirationDateCustom" type="datetime-local" name="expirationDate" 
+                            [(ngModel)]="expirationDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
```
MGibson1, line 85: same for this date picker

MGibson1 (commented):
Just some thoughts on disable Send policy

src/app/send/add-edit.component.html
addisonbeck: Disable Send hasn't been considered much yet so far (it's in the "not yet ready" list). Some inputs make use of `disableSend` just from how they were added to the form (whether I wrote them out manually or copy/pasted & adjusted), but that is something I'll implement fully soon another PR.

src/app/send/add-edit.component.html
addisonbeck: Addressed above

src/app/send/add-edit.component.html
addisonbeck: Addressed above

src/app/send/add-edit.component.html
MGibson1: sick thanks!

addisonbeck:
Closing this as I am nearing completion of everything else

PR #724, opened 2021-02-18
Finished Send Implementation

addisonbeck:
Follow up to https://github.com/bitwarden/desktop/pull/722

### jslib
- update to latest

### app.component
- Removed some unused imports & services (angular2google, nativeMessaging)
- Added several subscriptions that previously existed in the vault.component. Some of these were 1:1 copy/paste from one component to another (premium required, syncVault, checkSyncVault), others required  varying degrees of new implementations (new{CipherType}, newFolder, openPasswordGenerator).

### add-edit.component
- Cleared `password` on refresh since we don't ever show that value after creation
- Removed the `<ng-container>` wrapping the form and moved its `*ngIf` to the form's `.inner-content` so the page layout doesn't flicker in and out of existence when changing operations. 
- Added support for creating Sends within the form (create specific inputs, etc.).
- Added support for a cancel button that clears the form & sets its action to `None`. This works the same way the Vault cancel button does.
- Added "Share on Save"
- Added "Copy Send URL" button to the footer
- Added a "Show Password" button
- Hid optional fields by default with an expand/collapse
- Implemented the Disable Send policy
- Blocked creating File sends behind a Premium subscription

**NOTE:** The File/Text switcher is a radio button. Mobile & (I think) browser are using a bootstrap toggle switcher. Because `web` currently is not using a toggle, and bootstrap isn't used on desktop, desktop is currently using radio buttons.

### send.component
- Added `onCancel`, `onSaveSend` and `onDeletedSend` output handlers to refresh the screen with relevant data from those actions.
- Added sync support. This was in jslib, but was removed for client specific logic elsewhere. The implementation here is exactly as it was previously in jslib.
- Returned from `selectSend()` if the clicked Send is already selected to prevent the screen from trying to refresh unnecessarily. 
- Added a right click menu for sends with an option to copy the Send URL and an option to delete the selected Send

### messages.json
- Added more strings, all of which existed already in the web client.

### box.scss
- Added an exception to the default `.box label` styling because it does not work with radio buttons. This should probably get a genuine refactor at some point, but the changes required seemed out of scope for the Send implementation.
- Added `radio` type exceptions to any rules that already had `checkbox` type exceptions
- Added radio button stylings in the form of `.box-content-row-radio`, similar to how other input type rows get styled.

### Completed Functionality:
- Send is fully ready for testing at this stage
- All functions of the menu bar need some love from testing too

## Recorded Demo

https://user-images.githubusercontent.com/15897251/108770510-5f1fe700-7528-11eb-88de-6e3b9f42268b.mov

### DIFF
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -67,7 +67,7 @@
                         </select>
                         <div class="subtext">{{'deletionDateDesc' | i18n}}</div>
                     </div>
-                    <div class="box-content-row" *ngIf="deletionDateSelect === 0">
+                    <div class="box-content-row" *ngIf="deletionDateSelect === 0 && !editMode">
                         <input id="deletionDateCustom" type="datetime-local" name="deletionDate" 
                             [(ngModel)]="deletionDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
                     </div>
@@ -80,7 +80,7 @@
                         </select>
                         <div class="subtext">{{'expirationDateDesc' | i18n}}</div>
                     </div>
-                    <div class="box-content-row" *ngIf="expirationDateSelect === 0">
+                    <div class="box-content-row" *ngIf="expirationDateSelect === 0 && !editMode">
                         <input id="expirationDateCustom" type="datetime-local" name="expirationDate" 
                             [(ngModel)]="expirationDate" required placeholder="MM/DD/YYYY HH:MM AM/PM" [readOnly]="disableSend">
                     </div>
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -1,4 +1,3 @@
-<!-- Fix the menu bar -->
 <form (ngSubmit)="submit()" [appApiAction]="formPromise">
     <div class="content">
         <div class="inner-content" *ngIf="send">
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -1,106 +1,159 @@
-<ng-container *ngIf="send">
-    <form (ngSubmit)="submit()" [appApiAction]="formPromise">
-        <div class="content" *ngIf="send">
-            <div class="inner-content">
-                <div class="box">
-                    <div class="box-header">
-                        {{'editSend' | i18n}}
+<form (ngSubmit)="submit()" [appApiAction]="formPromise">
+    <div class="content">
+        <div class="inner-content" *ngIf="send">
+            <div class="box">
+                <div class="box-header">
+                    {{title}}
+                </div>
+                <div class="box-content">
+                    <div class="box-content-row" appBoxRow>
+                        <label for="name">{{'name' | i18n}}</label>
+                        <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus>
                     </div>
-                    <div class="box-content">
-                        <div class="box-content-row" appBoxRow>
-                            <label for="name">{{'name' | i18n}}</label>
-                            <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus>
+                </div>
+            </div>
+            <div class="box">
+                <div class="box-content">
+                    <ng-container *ngIf="!editMode">
+                        <div class="box-content-row box-content-row-radio">
+                            <label class="radio-header">{{'whatTypeOfSend' | i18n}}</label>
+                            <div class="item" *ngFor="let o of typeOptions">
+                                <input type="radio" class="radio" [(ngModel)]="send.type" name="Type_{{o.value}}"
+                                    id="type_{{o.value}}" [value]="o.value" (change)="typeChanged(o)"
+                                    [checked]="send.type === o.value">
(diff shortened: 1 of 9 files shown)


src/app/send/add-edit.component.ts
```diff
@@ -26,11 +26,21 @@ export class AddEditComponent extends BaseAddEditComponent {
...
         const send = await this.loadSend();
         this.send = await send.decrypt();
         this.hasPassword = this.send.password != null && this.send.password.trim() !== '';
         this.deletionDate = this.dateToString(this.send.deletionDate);
         this.expirationDate = this.dateToString(this.send.expirationDate);
-        this.password = null;
+    }
+
+    cancel() {
+        this.onCancelled.emit(this.send);
```
cscharf, line 38: Do you need to check for `ths.onCancelled != null`? Not sure, maybe not, but if the delegate method isn't set could throw an exception here.

src/scss/box.scss
```diff
@@ -7,6 +7,7 @@
     .box-header {
         margin: 0 10px 5px 10px;
         text-transform: uppercase;
+        display: flex;
```
cscharf, line 10: 💪🏻  flex!

src/app/send/add-edit.component.html
```diff
@@ -1,67 +1,130 @@
...
+            <div class="box">
+                <div class="box-content">
+                    <ng-container *ngIf="!editMode">
+                        <div class="box-content-row box-content-row-radio">
+                            <label class="radio-header">{{'whatTypeOfSend' | i18n}}</label>
+                            <div class="item" *ngFor="let o of typeOptions">
+                                <input type="radio" class="radio" [(ngModel)]="send.type" name="Type_{{o.value}}"
+                                    id="type_{{o.value}}" [value]="o.value" (change)="typeChanged(o)"
+                                    [checked]="send.type === o.value">
+                                <label class="label-hack" for="type_{{o.value}}">
```
cscharf, line 29: Is there are more comely name we can choose for this class other than `"label-hack"`? What is the intent of this class; I see the `:not()` rule in the stylesheet, however what is it not supposed to be doing? Let's name it explicitly based on function vs. what it is (a hack) 🪓

src/scss/box.scss
```diff
@@ -118,7 +128,7 @@
         overflow-x: auto;
     }
 
-    .row-label, label {
+    .row-label, label:not(.label-hack) {
```
cscharf, line 131: Let's rename our "`.label-hack` in a `box`" 😏

src/app/send/add-edit.component.html
addisonbeck: Unfortunately I'm not sure what that functional name would be. Maybe `.unstyled-label`? All applying this class is doing is ignoring the `<label>` styling that is forced on it just for being in a `.box-row`. Things like `display: block; width: 100%` aren't normal `<label>` styles and don't work in the context of radio buttons, but there is no way around that as long as all `<label>`s are getting that style. 

The real solution here would be to remove the default `label` stylings and use `.row-label` when we need these styles, so I guess my hope with just calling the class `label-hack` is that it will keep attention and get removed. Checking every label in the app just seemed out of scope for Send, especially for testing.

cscharf: Great thoughts, thanks. how about `unstyled` and you can put some comments in the scss that relate to the HACK and TODO nature of what you have mentioned above.

src/app/send/add-edit.component.ts
addisonbeck: onCancelled has a default emitter [here](https://github.com/bitwarden/jslib/blob/8a3b551c6e4113035dc40a379f43f8778d208ffe/src/angular/components/send/add-edit.component.ts#L34), and that's all that this call uses.

src/app/send/add-edit.component.html
addisonbeck: Done!

src/scss/box.scss
addisonbeck: Addressed above

cscharf (commented):
Awesome, let me know when you get the menu bar working and this is ready for review!

src/app/send/add-edit.component.html
```diff
@@ -1,67 +1,130 @@
...
-                            <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus>
+                </div>
+            </div>
+            <div class="box">
+                <div class="box-content">
+                    <ng-container *ngIf="!editMode">
+                        <div class="box-content-row box-content-row-radio">
+                            <label class="radio-header">{{'whatTypeOfSend' | i18n}}</label>
+                            <div class="item" *ngFor="let o of typeOptions">
+                                <input type="radio" class="radio" [(ngModel)]="send.type" name="Type_{{o.value}}"
```
cscharf, line 26: @addisonbeck , is there a way to use radio-buttons here (toggle) so it matches mobile and web? Browser should be getting the same treatment as well.

addisonbeck: @cscharf latest web appears to just be using normal radio buttons. Is that something else we need to update there?

cscharf (approved):

PR #730, opened 2021-02-23
Send Cleanup

addisonbeck:
There were a few issues I noticed while recording the demo video for desktop Send that never got committed for the PR.

1. Changed a few [disableSend] rules, disabling radio buttons & removing readOnly attributes that weren't doing anything. 
2. Also added a string and updated jslib.


cscharf (approved):
good catch

PR #734, opened 2021-02-24
Updated year to current in copyright(s)

djsmith85:
We all would like to forget 2020 for obvious reasons...


cscharf (approved):
Indeed, lol. Thanks!

PR #757, opened 2021-03-08
fix: Edge supported by biometrics

aimuz:
reference resources: https://docs.microsoft.com/en-us/microsoft-edge/extensions-chromium/developer-guide/native-messaging?tabs=macos

### DIFF
--- a/resources/entitlements.mas.plist
+++ b/resources/entitlements.mas.plist
@@ -12,6 +12,7 @@
 	<array>
 		<string>/Library/Application Support/Mozilla/NativeMessagingHosts/</string>
 		<string>/Library/Application Support/Google/Chrome/NativeMessagingHosts/</string>
+		<string>/Library/Application Support/Microsoft Edge/NativeMessagingHosts/</string>
 	</array>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


Hinton (dismissed):

Hinton:
@aimuz Did you have issues using browser biometrics on Edge before making these changes?

On Windows the `Computer\HKEY_CURRENT_USER\SOFTWARE\Google\Chrome` registry directory seems to always exist even on a fresh install without Chrome. Since Edge seems to fallback to this path things should work fine on Windows.

It might be different on Mac and Linux though, so to be safe these changes should ensure we always have the manifest available even if Chrome is not installed. 👍

aimuz:
I've been tested on MacOS and it works. I don't have a biometric Linux device, but I think it works.According to the edge document, there should be no risk of such a change

If it works on Windows itself, I will return the changes to windows

aimuz:
If you have the relevant equipment, I hope you can help me

Hinton (approved):

Hinton:
Confirmed Edge works on a clean Windows VM without chrome on currently released version.

aimuz:
Very good. When can we merge?

cscharf:
Please note, while merged this will not be released in the currently planned release as we've already cut our release candidates and are wrapping up regression testing, but should be in the next release.

PR #767, opened 2021-03-10
style tweaks to send add/edit page

kspearrin:
Looks like we introduced a new style (`subtext`) to accommodate the UI help text for the Send add/edit page. This wasn't necessary since we can use the existing `help-block` (see example on the settings page). Additionally, using help-block text is not desirable if we can use a box-footer instead. Sometimes that isn't possible since we have multiple rows we would like to annotate with help-blocks (for example, deletion and expiration dates). In this case, we can use the help-block.

## Screenshots

Before

image

After

image

### DIFF
--- a/src/app/send/add-edit.component.html
+++ b/src/app/send/add-edit.component.html
@@ -15,45 +15,43 @@
                         <label for="name">{{'name' | i18n}}</label>
                         <input id="name" type="text" name="Name" [(ngModel)]="send.name" appAutofocus [readOnly]="disableSend">
                     </div>
+                    <div class="box-content-row box-content-row-radio" *ngIf="!editMode">
+                        <label class="radio-header">{{'whatTypeOfSend' | i18n}}</label>
+                        <div class="item" *ngFor="let o of typeOptions">
+                            <input type="radio" class="radio" [(ngModel)]="send.type" name="Type_{{o.value}}"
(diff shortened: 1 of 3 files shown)


src/app/send/add-edit.component.html
```diff
@@ -15,45 +15,43 @@
...
-                            <label for="text">{{'text' | i18n}}</label>
-                            <textarea id="text" name="text" [(ngModel)]="send.text.text" rows="6" [readOnly]="disableSend"></textarea>
-                            <div class="subtext">{{'sendTextDesc' | i18n}}</div>
-                        </div>
-                        <div class="box-content-row box-content-row-checkbox" appBoxRow *ngIf="send.type === sendType.Text">
-                            <label for="hideText">{{'textHiddenByDefault' | i18n}}</label>
-                            <input id="hideText" name="hideText" type="checkbox" [(ngModel)]="send.text.hidden" [disabled]="disableSend">
-                        </div>
-                    </ng-container>
+                    <div class="box-content-row box-content-row-checkbox" appBoxRow *ngIf="send.type === sendType.Text">
```
cscharf, line 51: since this is nested in a `<div/>` that already has `*ngIf="send.type === sendType.Text"`, this is redundant here.

cscharf (approved):

PR #779, opened 2021-03-15
support vivaldi for macos

aimuz:
fix https://github.com/bitwarden/browser/issues/1590

### DIFF
--- a/src/main/nativeMessaging.main.ts
+++ b/src/main/nativeMessaging.main.ts
@@ -87,22 +87,18 @@ export class NativeMessagingMain {
                 this.createWindowsRegistry('HKCU\\SOFTWARE\\Google\\Chrome', 'HKCU\\SOFTWARE\\Google\\Chrome\\NativeMessagingHosts\\com.8bit.bitwarden', path.join(destination, 'chrome.json'));
                 break;
             case 'darwin':
-                if (existsSync(`${this.homedir()}/Library/Application\ Support/Mozilla/NativeMessagingHosts/`)) {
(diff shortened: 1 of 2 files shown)


Hinton:
@aimuz We will need to add the file to the `entitlements.mas.plist` in order to have access to actually write the file.

src/main/nativeMessaging.main.ts
```diff
@@ -87,22 +87,24 @@ export class NativeMessagingMain {
...
-                    this.writeManifest(`${this.homedir()}/Library/Application\ Support/Google/Chrome/NativeMessagingHosts/com.8bit.bitwarden.json`, chromeJson);
-                } else {
-                    this.logService.warning(`Chrome not found skipping.`);
-                }
-
-                if (existsSync(`${this.homedir()}/Library/Application\ Support/Microsoft\ Edge/NativeMessagingHosts`)) {
-                    this.writeManifest(`${this.homedir()}/Library/Application\ Support/Microsoft\ Edge/NativeMessagingHosts/com.8bit.bitwarden.json`, chromeJson);
-                } else {
-                    this.logService.warning(`Microsoft Edge not found skipping.`);
+                const nmhs = {
```
Hinton, line 90: Nice refactor, I believe we can take it a step further and extract it to a function or field to avoid duplicating it in the install and uninstall parts.

aimuz: Yes, it's a good idea. I missed that

Hinton (approved):

Issue #792, opened 2021-03-17
m1 osx / darwin universal or arm specific binary


nitrocode:
## Describe the Bug

It would be nice if the bitwarden/desktop app had a mac m1 binary.

Source: https://github.com/electron-userland/electron-builder/issues/5392#issuecomment-754239349

Looks like it can be done with a `--universal` switch added to `electron-builder`

https://github.com/bitwarden/desktop/blob/6873a6ea397002eb286b646704fa692950fb5a11/package.json#L40-L42

PR #793, opened 2021-03-17
Add --universal to mac builds

nitrocode:
Closes https://github.com/bitwarden/desktop/issues/792


package.json
```diff
@@ -37,9 +37,9 @@
     "clean:l10n": "git push origin --delete l10n_master",
     "pack:dir": "npm run clean:dist && electron-builder --dir -p never",
     "pack:lin": "npm run clean:dist && electron-builder --linux --x64 -p never",
-    "pack:mac": "npm run clean:dist && electron-builder --mac -p never",
-    "pack:mac:mas": "npm run clean:dist && electron-builder --mac mas -p never",
-    "pack:mac:masdev": "npm run clean:dist && electron-builder --mac mas-dev -p never",
+    "pack:mac": "npm run clean:dist && electron-builder --mac --universal -p never",
```
Hinton, line 40: This unfortunately doubles the application size, we can't do much about the Mac App Store unfortunately but for `dmg` we should strive towards creating two separate binaries which the users may download.

nitrocode: Thank you for the comment. I added a `pack:mac_arm64` target and reverted the `pack:mac` target.

I left the `--universal` target for the Mac App Store.

package.json
```diff
@@ -38,8 +38,9 @@
     "pack:dir": "npm run clean:dist && electron-builder --dir -p never",
     "pack:lin": "npm run clean:dist && electron-builder --linux --x64 -p never",
     "pack:mac": "npm run clean:dist && electron-builder --mac -p never",
-    "pack:mac:mas": "npm run clean:dist && electron-builder --mac mas -p never",
-    "pack:mac:masdev": "npm run clean:dist && electron-builder --mac mas-dev -p never",
+    "pack:mac_arm64": "npm run clean:dist && electron-builder --mac --arm64 -p never",
```
Hinton, line 41: Nitpick but for the other jobs we use `:`, not `_`. @joseph-flinn Something you could fix while hooking up the CI jobs perhaps?

Hinton (approved):

Hinton:
Looks good, thanks for the contribution @nitrocode.

package.json
cscharf: I would prefer we keep consistency here across the board. @nitrocode , can you please fix this syntax?
```suggestion
    "pack:mac:arm64": "npm run clean:dist && electron-builder --mac --arm64 -p never",
```

cscharf (changes requested):

nitrocode:
@Hinton I kept the target consistent using @cscharf suggestion. Thank you!

cscharf (approved):
Thank you!

PR #803, opened 2021-03-21
Patch electron-builder to support universal binaries on Mac

Hinton:
## Overview
When building the Mac App Store package we produce a universal binary (https://github.com/bitwarden/desktop/pull/793.). Unfortunately we get an error when we attempt to include our Safari extension, since it has already been compiled as a universal-binary. Compiling two separate binaries for the extension likewise produces another error. There is an open issue in electron-builder, https://github.com/electron-userland/electron-builder/issues/5552.

However in the meantime this PR will patch electron-builder using [patch-package](https://www.npmjs.com/package/patch-package) to handle our use case.


cscharf (approved):

cscharf:
@Hinton , please add a card to Asana for tech-debt cleanup of this item once upstream issues are resolved/merged in and we can update electron-builder accordingly.

PR #814, opened 2021-03-25
Simplify logic for resolving path to exe for native messaging

Hinton:
## Overview
Our hardcoded path on Mac makes it difficult to QA test Native messaging now that masdev works. This PR changes it to use `app.GetPath('exe')`. I've tested it in in devtools on our current Mac App Store release and it returns the expected path.

### Testing Considerations
We need to do regression testing on Mac and Windows (linux doesn't support browser biometrics yet) to ensure nothing broke by changing the path.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -119,7 +119,7 @@ export class Main {
             this.biometricMain = new BiometricDarwinMain(this.storageService, this.i18nService);
         }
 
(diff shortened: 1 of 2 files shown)


cscharf (approved):
@Hinton , thanks, please ensure there's a card in Asana for the regression testing of this so the QA team has a head's up and can plan it accordingly.

PR #816, opened 2021-03-29
Update macOS icon to follow Big Sur design

### LINKED ISSUE #788: Icon shape on macOS doesn't follow macOS Big Sur guidelines
## Describe the Bug

Icon shape does not match macOS Big Sur size and shape.

## Steps To Reproduce

1. Install Bitwarden for Mac
2. Open it
3. Look at it on the dock

## Expected Result

Icon size and shape should match.

## Actual Result

Icon size and shape doesn't match.

## Screenshots or Videos

image

## Environment

- Operating system: macOS 11.2.2 Big Sur, M1
- Build Version (go to "Settings" → "About"): Couldn't find build version in Bitwarden app

LaurentTreguier:
Fixes #788.

The icon was made using Apple's official icon template, simply changing the default background to the same gradient as the iOS icon in https://github.com/bitwarden/mobile/pull/962.
The shield also gets a gradient going from `#FFFFFF` to `#CFCFCF` as well as a slight shadow (opacity at 35%, blur = 5% of the icon's total height, and Y offset = 2.5% of the icon's total height).
Those values are completely arbitrary, icons seem to not always have the same values for things like shadow blur and such. So I tried going for something that "feels right" according to my own (very subjective) opinion.

Here is a quick visual comparison of the result on the dock:
image
image

Here is a Figma export for reference/tweaking:
[Bitwarden Desktop Icon.fig.zip](https://github.com/bitwarden/desktop/files/6223846/Bitwarden.Desktop.Icon.fig.zip)


mishushakov (dismissed):

tgreer-bw (changes requested):
The padding on the 16x16 and 32x32 images seems to be too much. Can They be adjusted to fit the whole 16x and 32x? If using an ICNS generator,  you may need to manually create the files at those sizes and bundle them into the ICNS as well.

LaurentTreguier:
> The padding on the 16x16 and 32x32 images seems to be too much.

Do you mean the padding of the shield inside the background, or the padding of the whole icon within its 16x16/32x32 bounds ?
I'm guessing it's the former; at least this is where I see a difference compared the other sizes

tgreer-bw:
Sorry for the delay! 

Yes, I meant that the shield should be closer to the size of the image with almost no padding on those 16x, 16@2x, 32x, etc. to match as closely to the prior icons as possible.

Thank you again for your contribution!

tgreer-bw:
I'm sorry, I re-read my reply and I was not clear at all. The larger shield is good, but the entire icon/blank padding should be almost the whole image size, like the originals.

LaurentTreguier:
Hmm. Now I'm getting confused with this... Did you actually mean that padding around both the shield and its background ?
On the 16x16 icon, the shield + blue background is 14x14, with a 1px gap all around it; is this 1px gap the padding you are meaning to remove ?

tgreer-bw:
Correct, the transparent padding around the image is what should be removed.

LaurentTreguier:
That padding should probably not be removed; it's part of the macOS Big Sur icon styling. Big Sur icons always have a padding around them; the padding is used to give them a slight shadow, and some icons also use it to fit additional items over the icon:
image
The padding is intentional, it's part of Apple's current guidelines; all icon sizes have it in their official icon template, even down to the 16x16 and 32x32 icons. While this padding makes the icon differ a bit from the original, it's necessary to make it fully consistent with other macOS icons, which is the main goal of this icon change.

tgreer-bw (approved):

cscharf (approved):

PR #819, opened 2021-03-31
Fix Premium Required message after sso auth

eliykat:
## Objective
Users were getting a "Premium required" error message when trying to login via SSO, even though they were members of a paid organization.

## Code changes
Put in a happy little `break` statement. The `ssoCallback` case was flowing into the `premiumRequired` case, which happened to be immediately after.


cscharf (approved):
Nice catch!

PR #830, opened 2021-04-07
Remove electron remote

Hinton:
## Objective
Since Electron has [deprecated the remote module](https://www.electronjs.org/docs/api/remote) we need to migrate away from it. The recommended alternative is to use `ipc` to explicitly communicate between the main and renderer.

### Testing Considerations
General regression testing to verify nothing broke.

Depends on https://github.com/bitwarden/jslib/pull/332

### DIFF
--- a/src/app/send/send.component.ts
+++ b/src/app/send/send.component.ts
@@ -1,5 +1,3 @@
-import { remote } from 'electron';
-
 import {
     Component,
     NgZone,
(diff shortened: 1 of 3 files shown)


cscharf (approved):

cscharf (changes requested):
Update of `jslib` that includes https://github.com/bitwarden/jslib/pull/330 will require an update to the constructor of the i18n service. The build has failed.

PR #838, opened 2021-04-14
Password reprompt

Hinton:
## Objective
Adds support for requiring master password re-prompt before accessing hidden fields on ciphers. Also marks card numbers as a secret to ensure it's protected.

### Code Changes
- **src/app/vault/add-edit.component.html**: Hide card number, add prompt setting.
- **src/app/vault/vault.component.ts**: Prevent copying, using context menu on vault page.
- **src/app/vault/view.component.html**: Hide card number.

### Testing Considerations
We should verify all expected values are protected and cannot be accessed without prompting for master password.

### Screenshots
electron_002XftzBMk
electron_jpVi7LoVYb
electron_kzHJoLy43n

Depends on https://github.com/bitwarden/jslib/pull/343


src/app/vault/vault.component.ts
```diff
@@ -638,8 +644,13 @@ export class VaultComponent implements OnInit, OnDestroy {
         this.functionWithChangeDetection(() => this.addCipher(type));
     }
 
-    private copyValue(value: string, labelI18nKey: string) {
-        this.functionWithChangeDetection(() => {
+    private copyValue(cipher: CipherView, value: string, labelI18nKey: string, aType: string) {
+        this.functionWithChangeDetection(async () => {
+            if (cipher.reprompt !== CipherRepromptType.None && this.passwordRepromptService.protectedFields().includes(aType) &&
+               ! await this.passwordRepromptService.showPasswordPrompt()) {
```
cscharf, line 650: Remove space between bang (`!`) and `await` keyword

src/app/vault/vault.component.ts
```diff
@@ -342,6 +346,8 @@ export class VaultComponent implements OnInit, OnDestroy {
             return;
         } else if (this.dirtyInput() && await this.wantsToSaveChanges()) {
             return;
+        } else if (cipher.reprompt !== CipherRepromptType.None && ! await this.passwordRepromptService.showPasswordPrompt()) {
```
cscharf, line 349: Remove space between bang (`!`) and `await` keyword

src/app/vault/vault.component.ts
```diff
@@ -330,6 +332,8 @@ export class VaultComponent implements OnInit, OnDestroy {
...
```
cscharf, line 335: Remove space between bang (`!`) and `await` keyword

src/app/vault/vault.component.ts
```diff
@@ -638,8 +644,13 @@ export class VaultComponent implements OnInit, OnDestroy {
         this.functionWithChangeDetection(() => this.addCipher(type));
     }
 
-    private copyValue(value: string, labelI18nKey: string) {
-        this.functionWithChangeDetection(() => {
+    private copyValue(cipher: CipherView, value: string, labelI18nKey: string, aType: string) {
+        this.functionWithChangeDetection(async () => {
+            if (cipher.reprompt !== CipherRepromptType.None && this.passwordRepromptService.protectedFields().includes(aType) &&
```
cscharf, line 649: I don't see a change in state for `reprompt` nor do I see any internal state tracking for the `platformUtilsService` implementations or the `passwordRepromptService` which manages whether a reprompt for a given item has already been satisfied for a given "view" of that item. Is it the intent that for every literal copy operation, even after re-entering my MP for the intial one for a given item, I would continue to be re-prompted? So if I click `copy password` 10 times in a row while viewing that item, I would be prompted for my MP 10 times? If that's the intended behavior, cool (just need to document that I guess and we'll see how it rolls in real-world usage scenarios).

cscharf (commented):
Looks good overall, just a few comments.

src/app/vault/vault.component.ts
Hinton: @cscharf It's slightly tricky to keep the state in vault.component since it's global for all items in the list. This is only for the right click action, if you enter the cipher and do the actions that way it will only prompt once. We could explore having a dict with the timestamp and some threshold but it feels slightly complicated for the initial version.

cscharf (approved):

PR #844, opened 2021-04-19
Add custom-pipe to correct an issue with display of Unicode/Emoji passwords #768

gryffs:
Hi, I've looked into issue #768 and the problem is associated with an angular pipe `ColorPasswordPipe` used for sanitizing HTML and colorizing the password. Unicode characters do not play well with this Pipe because of issues with Unicode, ie: `"💩".length === 2`, and how it parsed through a basic string. I've resolved the issue by creating a custom-pipe that can check for Unicode / Emoji characters. Unicode is an interesting rabbit hole to go down 😃 

Just because I can does not mean I should, please let me know if I should not have gone this route, one custom pipe problem is now we have to maintain it. I've admittedly had a hard time trying to find the source of the original `ColorPasswordPipe` but I figured I could submit this and start a discussion. The benefit to allowing Emoji in passwords (so Bitwarden supporting the correct view of them) is that they do create more complex passwords for users.

I set the new custom pipe `ColorPasswordPipeWithEmoji` to replace the Angular Pipe mentioned above in all locations being used.

The new pipe is simply an updated version of the Angular Pipe here: [ColorPasswordPipe](https://gist.github.com/gryffs/d171c8e6ad07ef5c1ce73a2566d2dbff)

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -43,9 +43,11 @@ import { StopClickDirective } from 'jslib/angular/directives/stop-click.directiv
 import { StopPropDirective } from 'jslib/angular/directives/stop-prop.directive';
 import { TrueFalseValueDirective } from 'jslib/angular/directives/true-false-value.directive';
 
-import { ColorPasswordPipeWithEmoji } from '../custom-pipes/color-password-with-emoji';
 import { I18nPipe } from 'jslib/angular/pipes/i18n.pipe';
 import { SearchCiphersPipe } from 'jslib/angular/pipes/search-ciphers.pipe';
+
+// custom-pipes
+import { ColorPasswordPipeWithEmoji } from '../custom-pipes/color-password-with-emoji';
 
 import { AddEditComponent } from './vault/add-edit.component';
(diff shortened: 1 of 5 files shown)


cscharf:
Hi @gryffs (awesome name btw), thanks for the contribution, however I'm wondering if it wouldn't be more prudent to just update the current color password pipe already implemented in jslib, therefore making this available across the other repos, w/o other changes to references or implementation(s).

see: https://github.com/bitwarden/jslib/blob/master/src/angular/pipes/color-password.pipe.ts

`jslib` is pulled in as a sub-module to all of our Angular TypeScript repos/clients (Web, Desktop, Browser), with this pipe likely used across Web, Desktop and Browser.

gryffs:
Great, thank you for the feedback @cscharf (Your name rocks too!). I probably should have asked, but thank you for pointing me to the jslib. This answers many of my unasked questions. Also, I did overlook the fact that the issue reported is across platforms so correcting it there should rectify all. I've forked the jslib and will update the pipe and submit again. I'll close this PR and reference once I get the new PR submitteed.

PR #853, opened 2021-04-23
Don't submit form when unlocking with biometrics

eliykat:
## Objective

PR #846 changed the "Unlock with biometrics" button to use `<button>` instead of `<a>`. However, this has the effect of submitting the form when the button is pressed, which prompts a "Master Password is required" error toast after logging in.

This will need to be cherry-picked into `rc` once merged.

## Code changes

Add `type="button"`, which prevents the button from submitting the form.

### DIFF
--- a/src/app/accounts/lock.component.html
+++ b/src/app/accounts/lock.component.html
@@ -38,7 +38,7 @@
                 </button>
             </div>
             <div class="buttons-row" *ngIf="supportsBiometric && biometricLock">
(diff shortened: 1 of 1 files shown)


Hinton (approved):
Nice catch. I don't believe the other PR is in the `rc` branch though but perhaps they should be.

eliykat:
Good point, I assumed it was in `rc` branch but it's not. I assume it's not worth loading up QA further though.

PR #860, opened 2021-04-28
Bump jslib to include fix for lastpass totp import

djsmith85:
Bumping jslib to include a fix for the missing totp secret in a lastpass import done in bitwarden/jslib#361


cscharf (approved):

cscharf:
Thanks!

PR #855, opened 2021-04-23
Vertically align the name and icon in secured notes list.

gryffs:
Hello, this PR is meant to correct the issue here: #599

The above issue was confirmed on mac as well.

I've aded the classes `box-content-row box-content-row-flex` in order to align the text with the image.

Issue #599, opened 2020-11-26
Secure note label misaligned with note icon


gprst:
## Steps To Reproduce

Go to the Secure Notes menu of your vault.

## Expected Result

The name of the note should be vertically aligned with the note icon, as for the web extension.

## Actual Result

The name of the note is misaligned with the note icon (slightly upward).

## Screenshots or Videos

![misaligned](https://user-images.githubusercontent.com/32643653/100368777-4a54b580-3004-11eb-8d08-b832487011c4.png)

## Environment

- Operating systems: Windows 10
- Build Version : 1.22.2

djsmith85:
Closing this, as it was solved with https://github.com/bitwarden/desktop/pull/870

Thank you again @gryffs for your work in this.

PR #870, opened 2021-05-03
Add flexbox to cipher list to correct display center issue.

gryffs:
This is a better way to correct the issue in #599 
First pull request I made for this issue was #855 But I closed it because it had side effects.

I've added an additional `<a>` wrapper to implement better flexbox functionality so that the items in question can consistency be centered. This also required adding an additional scss field fo referenced wrapper.

Before: 
Screen Shot 2021-05-03 at 4 33 30 PMScreen Shot 2021-05-03 at 4 34 13 PM

After:
Screen Shot 2021-05-03 at 4 39 29 PMScreen Shot 2021-05-03 at 4 39 42 PM

I've confirmed that the side effects mentioned in the previous pull request have been corrected. The "After" screenshot above shows the list display of all types (I could not add an attachment to confirm placement, but share notification is there).

Let me know what you think, hope this helps.

### DIFF
--- a/src/app/vault/ciphers.component.html
+++ b/src/app/vault/ciphers.component.html
@@ -12,20 +12,22 @@
             (scrolled)="loadMore()">
             <a *ngFor="let c of filteredCiphers" appStopClick (click)="selectCipher(c)"
                 (contextmenu)="rightClickCipher(c)" href="#" title="{{'viewItem' | i18n}}"
-                [ngClass]="{'active': c.id === activeCipherId}">
+                [ngClass]="{'active': c.id === activeCipherId}" class="flex-list-item">
                 <app-vault-icon [cipher]="c"></app-vault-icon>
-                <span class="text">
-                    {{c.name}}
-                    <ng-container *ngIf="c.organizationId">
-                        <i class="fa fa-share-alt text-muted" title="{{'shared' | i18n}}" aria-hidden="true"></i>
-                        <span class="sr-only">{{'shared' | i18n}}</span>
-                    </ng-container>
-                    <ng-container *ngIf="c.hasAttachments">
-                        <i class="fa fa-paperclip text-muted" title="{{'attachments' | i18n}}" aria-hidden="true"></i>
-                        <span class="sr-only">{{'attachments' | i18n}}</span>
-                    </ng-container>
-                </span>
-                <span class="detail">{{c.subTitle}}</span>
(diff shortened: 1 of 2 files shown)


src/app/vault/ciphers.component.html
```diff
@@ -12,20 +12,22 @@
...
-                        <i class="fa fa-share-alt text-muted" title="{{'shared' | i18n}}" aria-hidden="true"></i>
-                        <span class="sr-only">{{'shared' | i18n}}</span>
-                    </ng-container>
-                    <ng-container *ngIf="c.hasAttachments">
-                        <i class="fa fa-paperclip text-muted" title="{{'attachments' | i18n}}" aria-hidden="true"></i>
-                        <span class="sr-only">{{'attachments' | i18n}}</span>
-                    </ng-container>
-                </span>
-                <span class="detail">{{c.subTitle}}</span>
+                <a class='flex-cipher-list-item'>
```
cscharf, line 17: Please use double quotes to encapsulate HTML attributes.

cscharf: Also, does the `a > a` structure mess with the tab index or keyboard interaction (accessibility)? Just wondering if it wouldn't be more semantically correct to use a `<span>` or since HTML5 allows it, a `<div>`? No particular prejudice against `<a>` here, just want to ensure it's interactive via keyboard still.

cscharf:
Thanks @gryffs , looks good, just a couple of comments (one small syntax change and one question).

src/app/vault/ciphers.component.html
gryffs: Thanks and I see your point. I would probably opt for the `<div>` since `<span>` has a history of `display:inline`. Keyboard interaction looks to remain intact and accessibility is functioning the same as before as well.
Any thoughts around that before I push up these updates?

Thanks again for the feedback.

cscharf: Let's go with `<div>`, thanks!

gryffs:
Corrected a syntax issue and changed the above referenced `<a>` element to a `<div>` element. Also attached a small video showing tab through on list.


https://user-images.githubusercontent.com/16548994/116944228-a7d1ea00-aca7-11eb-821e-33cae61f930e.mov

cscharf (approved):
Thanks!

Issue #711, opened 2021-02-04
Windows Hello authentication prompt is displayed even when Bitwarden starts to tray


hyunbridge:
## Describe the Bug

Even when Bitwarden starts to tray, Windows Hello authentication prompt is displayed.

## Steps To Reproduce

1. Go to "File" → "Settings"
2. Enable "Start To Tray Icon" and "Start automatically on login"
3. Restart your system

## Expected Result

Windows Hello authentication prompt should not be displayed when Bitwarden starts to tray.

## Actual Result

Windows Hello authentication prompt is displayed.

## Screenshots or Videos

https://user-images.githubusercontent.com/51025374/106864434-35378980-670d-11eb-861e-da8ddeb511a0.mp4

## Environment

- Operating system: Windows 10 Pro (Build 19042.789)
- Build Version (go to "Settings" → "About"): 1.24.6

PR #873, opened 2021-05-06
Add setting for disabling auto prompt of biometrics

Hinton:
## Objective
When starting the desktop application on launch or when running using `npm run electron`, the biometrics prompts can be quite annoying. This PR adds a setting to the desktop application for disabling the auto prompt behaviour. It also verifies that the window is visible before prompting for biometrics.

### Code Changes
- **src/app/accounts/lock.component.ts**: Check if no auto prompt is enabled and if so do not prompt for biometrics automatically.
- **src/app/accounts/settings.component.html**: Html for toggle.
- **src/app/accounts/settings.component.ts**: Logic for toggling no auto prompt.
- **src/locales/en/messages.json**: Locale updates.

### Testing Considerations
We should do some quick regression testing to ensure current behavior is unchanged (prompt by default) and the setting works. When starting minimized, no biometric prompt should appear even when option is disabled. When disabling biometrics, this setting should also become disabled.

### Screenshots
image

Depends on https://github.com/bitwarden/jslib/pull/370
Resolves: https://github.com/bitwarden/desktop/issues/711

### DIFF
--- a/src/app/accounts/lock.component.ts
+++ b/src/app/accounts/lock.component.ts
@@ -7,6 +7,7 @@ import {
     ActivatedRoute,
     Router,
 } from '@angular/router';
+import { ipcRenderer } from 'electron';
 
 import { ApiService } from 'jslib/abstractions/api.service';
 import { CryptoService } from 'jslib/abstractions/crypto.service';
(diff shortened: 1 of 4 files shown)


Hinton:
@cscharf Depends on jslib.

cscharf (approved):

PR #897, opened 2021-05-14
Add missing languages from Crowdin

tgreer-bw:
Langauge updates for Desktop

PR #898, opened 2021-05-17
Fixed locale import for Montenegrin (Latin)

djsmith85:
With #897 an import for the Montenegrin locale was wrong, which let the build fail. Based on the two letter code used in Crowdin, I looked up the three letter lang code (**cnr**) on wikipedia and then searched/crossreferenced in [CLDR](https://github.com/unicode-org/cldr) which is used by Angular as source for the locales.


cscharf (approved):

cscharf:
thanks @djsmith85

cscharf:
Gah, missed this 🙁 

```
src/app/app.module.ts:94:1
ERROR: 94:1  ordered-imports  Import sources within a group must be alphabetized.
```

PR #164, opened 2018-11-19
Add .nvmrc to use node 10

libertylocked:
Now that the latest node version is v11, and `bitwarden-desktop` does not build on node 11, some users are experiencing build issues on Arch Linux.
With NVM we are able to build it with the correct node version. Some other electron apps use `.nvmrc` too like signal-desktop

PR #903, opened 2021-05-18
Update .nvmrc from 10.13.0 to 14.17.0

djsmith85:
With PR #164 a .`nvmrc` file was introduced to make it easier to choose the right version of node. As we've recently migrated to Node 14 LTS, I've updated the `.nvmrc` accordingly


cscharf (approved):
Thanks @djsmith85

PR #904, opened 2021-05-18
Add precommit githook and execute npm run lint

djsmith85:
Recently I issued a [PR](https://github.com/bitwarden/desktop/pull/898) to fix a broken build which got merged, but unfortunately I had forgotten to run `npm run lint:fix` before committing. This resulted in another broken build, as linting failed.

To prevent this happening with future commits, I have added a pre-commit hook which runs `npm run lint`.

While committing staged changes, the command `npm run lint` will be executed. If this throws any errors, committing is aborted.

This currently only checks for linting issues, and it's up to the committer to fix the issues manually. Up to discussion could be to run `npm run lint:fix` which would fix issues if possible and finish up the commit.

The following npm packages were used: `husky` and `lint-staged`

Please review this carefully and also check if changes are needed in regard to the CI build: https://typicode.github.io/husky/#/?id=disable-husky-in-cidocker


eliykat:
I think this depends on each dev's workflow - my code is definitely worse before it gets better and I generally run `lint` as the last thing I do, before pushing for review. Is there a pre-push-to-remote hook you can use instead?

cscharf:
Thanks @djsmith85 , I would actually prefer, after giving this some thought, that we simply add a PR check via GitHub actions that performs the `lint` run (relatively low risk and read-only action w/o any need for stored secrets, etc.) and have that be visible on the PR itself as a condition for merging. Therefore leaving it to the individual developer's own personal style and commit flow but giving PR reviewers and the submitter a final check before a merge is done. I'll add a card onto our DevOps board as they'll be tackling our build and deployment pipeline improvements pretty soon here.

djsmith85:
Thank you, @eliykat and @cscharf for the valuable feedback. My intention was to minimize build errors and fix-lint commits. I like the idea of having an additional PR check, which fixes the first issue and gives the committer more freedom in his/her workflow. It also shortens the feedback loop a little, as the check runs relatively fast and saves you (Bitwarden Team) some work in reviewing/merging multiple PRs. 

I'll close this PR. In the meantime I'll try and make it a habit to run linting before committing/pushing

PR #929, opened 2021-05-31
Respond to OS theme changes

### LINKED ISSUE #739: Default theme doesn't follow system theme changes
The app doesn't listen for changes to the system theme (sunrise and sunset for me):

https://github.com/bitwarden/desktop/blob/5d953fd2fc37d4e7c376d3e17f42df8eec16aa6e/src/app/services.module.ts#L157-L160

I assume it could be implemented like https://github.com/bitwarden/browser/pull/1373/files?

### LINKED ISSUE #896: Error: Cannot read property 'webContents' of null - when waking up laptop from sleep
## Describe the Bug

I get an alert with error message:

> Uncaught Exception:
TypeError: Cannot read property 'webContents' of null
at NativeTheme.<anonymous> (/Applications/Bitwarden.app/Contents/Resources/app.asar/main.js:19357:28)
at NativeTheme.emit (events.js:315:20)

## Steps To Reproduce

I cannot really reproduce, but I got the error after waking up my laptop from sleep

## Environment

- Operating system: Mac OS Big Sur
- Installation method: Homebrew
- Build Version 1.26.3 (526)

eliykat:
## Objective
* Fix #896, which was caused by a missing null check when calling `windowMain.win.webContents.send` when the system theme changed.
* Once I started looking at this, I found that this message was not caught by anything, and the theme was not updating when the system theme changed. Fix #739, which reported this issue.
* Then I had to consider whether to gate this for MacOS only, as we do when the theme is set initially [here](https://github.com/bitwarden/desktop/blob/c7345197f72be1f95870a089d04a27d2963987ef/src/app/services.module.ts#L159). I'm not sure why we do this, but the [documentation for nativeTheme](https://www.electronjs.org/docs/api/native-theme) doesn't say it's limited to MacOS, and it worked fine for Windows when I tested it. So I removed the MacOS gate when we set the theme initially as well. I'll make sure to mention this for QA; if it's fine here I can update browser as well for consistency.

## Code changes
* jslib: add null check - https://github.com/bitwarden/jslib/pull/397
* `app.component.ts`: use existing platformUtilsService method to catch and action "theme changed" message sent by main process
* `services.module.ts`: remove `if` condition so that all OS' can track the system theme

jslib needs to be merged & bumped first

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -124,6 +124,11 @@ export class AppComponent implements OnInit {
             window.onkeypress = () => this.recordActivity();
         });
 
+        this.platformUtilsService.onDefaultSystemThemeChange(theme => {
+            window.document.documentElement.classList.remove('theme_light', 'theme_dark');
+            window.document.documentElement.classList.add('theme_' + theme);
+        });
+
         this.broadcasterService.subscribe(BroadcasterSubscriptionId, async (message: any) => {
             this.ngZone.run(async () => {
                 switch (message.command) {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -124,14 +124,6 @@ export class AppComponent implements OnInit {
             window.onkeypress = () => this.recordActivity();
         });
 
-        this.platformUtilsService.onDefaultSystemThemeChange(async systemTheme => {
-            const theme = await this.storageService.get<string>(ConstantsService.themeKey);
-            if (theme == null) {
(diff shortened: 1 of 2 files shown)


Hinton:
I believe the reason it was macOS only is due to us previously using an older API for this which only supported macOS. The check was then never removed when upgrading electron.

eliykat:
Thanks for the context @Hinton. I just moved the theme update logic to be more consistent with how the browser client works, and bumped jslib, should be good to go now.

Hinton (approved):

PR #936, opened 2021-06-01
Update pipelines

vgrassia:
- Trigger `deploy` workflow manually
- Publish release draft at the end of the `deploy` workflow
- Enable artifact uploads (except MacOS) on all branches
- On asset upload, if asset is not found, error instead of warn

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,10 +7,9 @@ on:
       - 'gh-pages'
 
 jobs:
-
   cloc:
+    name: CLOC
     runs-on: ubuntu-latest
-
     steps:
       - name: Checkout repo
         uses: actions/checkout@5a4ac9002d0be2fb38bd78e4b4dbde5606d7042f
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -339,7 +340,7 @@ jobs:
         run: ./scripts/safari-build.ps1 -copyonly
 
       - name: Build application (dev)
-        if: github.ref != 'refs/heads/master' || github.ref == 'refs/heads/rc'
```
joseph-flinn, line 342: This should actually be `!=` (this is a part of that logic thing that I mentioned needs to be updated)

joseph-flinn (changes requested):

.github/workflows/build.yml
joseph-flinn: either that, or we can break out the logic of `npm run dist:mac:masdev` and remove that if statement altogether

vgrassia: Updated to have both be `!=`.

joseph-flinn (approved):

PR #941, opened 2021-06-03
Add jslib as a "real" dependency

Hinton:
## Objective
Importing jslib as a "real" dependency which provides an easier way to sync dependency versions across our projects. 

- Split jslib into multiple projects, desktop depends on `common`, `angular` and `electron`.
- Change from using `postinstall` to `preinstall` since we expect the jslib packages.json files.
- Change all import paths from `jslib` to `jslib-common`, `jslib-angular` or `jslib-electron`.

related to https://github.com/bitwarden/jslib/pull/363


src/app/dummy.module.ts
```diff
@@ -0,0 +1,14 @@
...
+import { SearchPipe } from 'jslib-angular/pipes/search.pipe';
+
+@NgModule({
+    imports: [],
+    declarations: [
+        InputVerbatimDirective,
+        SearchPipe,
+    ],
+})
+export class DummyModule {
```
MGibson1, line 13: What's this for?

Hinton: An old dummy module I removed a while back that seems to have re-appeared 😞 . (Probably during the merge conflicts). Removed again.

cscharf (approved):

PR #946, opened 2021-06-08
Authenticate with secure storage service

MGibson1:
# Overview

See bitwarden/jslib#402.

# Files Changed

* **main.ts**: Add biometric to keytarStorageListener. This is used to authenticate biometrically-secured keys
* **nativeMessagingService.ts**: Get key from cryptoService and give it to the requesting browser plugin. Biometric authentication is handled by the keytarListenerService.
* **webpack.main.ts**: Add source maps to the main process compilation. This enables debugging of the main process from `.ts` files rather than compiled `.js`

# Testing

Tests should be done around authentication with biometrics, pins, and passwords.

In particular, interaction between Never lock and biometric unlock is important.

There are upgrade considerations from a client which has either never lock or biometrics enabled. These clients should upgrade smoothly with no change in functionality.#946 

Authentication regression needs to be completed around the other clients as well (once they are upgraded), since authentication and unlocking code is shared across all jslib-dependent clients


webpack.main.js
```diff
@@ -28,7 +28,9 @@ const common = {
     output: {
         filename: '[name].js',
         path: path.resolve(__dirname, 'build'),
+        devtoolModuleFilenameTemplate: '[absolute-resource-path]',
     },
+    devtool: 'cheap-source-map'
```
Hinton, line 33: I don't think we'll want this devtools in production? We should either switch to a production ready one or toggle on/off depending on mode. https://webpack.js.org/configuration/devtool/

MGibson1: Good point. I'm thinking `source-map` since it matches our other clients. However, I'm not sure if these source maps are useful in production at all.

Does anyone know if you can attach to a production instance of an electron app's main process?

Hinton: https://stackoverflow.com/questions/45485262/how-to-debug-electron-production-binaries seems to be possible.

MGibson1: Per our discussion I went with building source maps only in dev builds

Hinton (approved):
LGTM, might want @bitwarden/dept-devops to check the updated CI flow.

joseph-flinn (approved):
Everything looks good to me. The real test on those caches will be to see if they are used in the other mac build jobs (they only run on master/rc).

PR #955, opened 2021-06-15
fix: radio instead of dropdown

knhash:
**Feature name:**

```
Replace the drop down for password/passphrase option with a radio button choice
```

**Feature Description**

There are only two options in the drop down, would it be better to replace it with a radio button for choice?
- Aids in discoverability: I completely missed the option for a whole year before I thought to click on the option and see what it contains.
- Less number of interactions: It would take 1 less click to get the same result

Community Forum post: https://community.bitwarden.com/t/radio-choice-password-passphrase/30636


src/app/vault/password-generator.component.html
```diff
@@ -27,13 +27,17 @@
...
-                            <label for="type" class="sr-only">{{'type' | i18n}}</label>
-                            <select id="type" [(ngModel)]="options.type" (change)="saveOptions()">
-                                <option value="password">{{'password' | i18n}}</option>
-                                <option value="passphrase">{{'passphrase' | i18n}}</option>
-                            </select>
-                        </div>
+                        <div class="box-content-row box-content-row-radio">
+                            <div class="radio-group text-default" appBoxRow name="PassTypeOptions"
+                                *ngFor="let o of passTypeOptions">
+                                <label for="type" class="sr-only">{{'type' | i18n}}</label>
```
eliykat, line 33: This `for` no longer refers to any element. Please delete the `for` attribute and add `radio-header` in the class. This is the pattern in the send add/edit dialog so we'll follow it here.

src/app/vault/password-generator.component.html
```diff
@@ -27,13 +27,17 @@
...
                     </div>
                     <div class="box-content condensed" [hidden]="!showOptions">
-                        <div class="box-content-row" appBoxRow>
-                            <label for="type" class="sr-only">{{'type' | i18n}}</label>
-                            <select id="type" [(ngModel)]="options.type" (change)="saveOptions()">
-                                <option value="password">{{'password' | i18n}}</option>
-                                <option value="passphrase">{{'passphrase' | i18n}}</option>
-                            </select>
-                        </div>
+                        <div class="box-content-row box-content-row-radio">
```
eliykat, line 30: I think this element is missing the closing `</div>` tag.

eliykat (changes requested):
Thanks for the PR! It also took me more than 6 months to find the "passphrase" option, so I think will be an improvement.

I've left a few comments which are hopefully self-explanatory, let me know if you have any questions.

eliykat:
I could find your other PR for browser, but not for [web](https://github.com/bitwarden/web). Please duplicate these changes for the web vault and tag me in the PR.

src/app/vault/password-generator.component.ts
```diff
@@ -13,8 +13,13 @@ import {
     templateUrl: 'password-generator.component.html',
 })
 export class PasswordGeneratorComponent extends BasePasswordGeneratorComponent {
+    passTypeOptions: any[];
```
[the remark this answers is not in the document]
knhash: Hey @eliykat . Thank you very much for your patience. All this is new to me so these were some silly mistakes, I appreciate your help.

I have raised the PR for `jslib` here: [JSLIB-PR](https://github.com/bitwarden/jslib/pull/420).

src/app/vault/password-generator.component.ts
```diff
@@ -13,8 +13,13 @@ import {
...
     constructor(passwordGenerationService: PasswordGenerationService, platformUtilsService: PlatformUtilsService,
         i18nService: I18nService) {
         super(passwordGenerationService, platformUtilsService, i18nService, window);
+        this.passTypeOptions = [
+            { name: 'password', value: 'password' },
```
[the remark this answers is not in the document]
knhash: Corrected in [JSLIB-PR](https://github.com/bitwarden/jslib/pull/420)

src/app/vault/password-generator.component.ts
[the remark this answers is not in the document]
eliykat: I don't think they're silly mistakes, they're all part of the normal code review process. :) Let me know when you've fixed everything and it's ready for me to look at again.

src/app/vault/password-generator.component.html
knhash: It was. Corrected.

src/app/vault/password-generator.component.html
knhash: Ack. I have removed the `for`. But I don't follow the radio header comment, there is already a title (OPTIONS) being set a little further up, is that sufficient?

src/app/vault/password-generator.component.ts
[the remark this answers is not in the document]
knhash: @eliykat The changes have been made. It is ready for your review.

knhash:
> I could find your other PR for browser, but not for [web](https://github.com/bitwarden/web). Please duplicate these changes for the web vault and tag me in the PR.

The web already has the option radio button style! Under Tools > Pass Generator.

Note: The pass generator while creating a new entry on web doesn't give options for the password.  But that is for another day.

src/app/vault/password-generator.component.html
```diff
@@ -27,12 +27,16 @@
...
-                                <option value="passphrase">{{'passphrase' | i18n}}</option>
-                            </select>
+                        <div class="box-content-row box-content-row-radio">
+                            <div class="radio-group text-default" appBoxRow name="PassTypeOptions"
+                                *ngFor="let o of passTypeOptions">
+                                <input type="radio" class="radio" [(ngModel)]="options.type" name="Type_{{o.value}}"
+                                    id="type_{{o.value}}" [value]="o.value" (change)="saveOptions()"
+                                    [checked]="options.type === o.value">
+                                <label class="unstyled" for="type_{{o.value}}">
+                                    {{o.name | i18n}}
```
eliykat, line 37: Ah, I didn't realize you'd done the internationalization here. Either way, now that we've moved the call to `i18nService` into the base class (in jslib), you can remove the `i18n` pipe here. So just: `{{o.name}}`.

src/app/vault/password-generator.component.html
```diff
@@ -27,12 +27,16 @@
                         </button>
                     </div>
                     <div class="box-content condensed" [hidden]="!showOptions">
-                        <div class="box-content-row" appBoxRow>
-                            <label for="type" class="sr-only">{{'type' | i18n}}</label>
```
eliykat, line 31: I didn't mean to get rid of this `<label>` altogether, we should keep it for screen readers. 
Just change it to:
`<label class="sr-only radio-header">{{'type' | i18n}}</label>`

eliykat (changes requested):
Just a couple of minor fixes and then it should be good to go.

I've merged the changes to jslib, so you can also update jslib while you're here.
1. delete jslib if you had it symlinked, then run `npm run sub:init`. If you didn't have it symlinked or don't know what that means, skip this step.
2. run `git submodule update --remote --reference upstream` - this will pull the latest version of the `jslib` submodule from github (with your changes in it)
3. now if you run `git diff`, you should see the change to the submodule commit hash. (Remember submodules are just a pointer to a commit hash on another repo.) You can check this against the latest commit hash on the `jslib` repo on Github. For example (your hashes will be different to this):
```bash
-Subproject commit b7acbcf3a7766eb0a52f21fe90505a69b4c19444
+Subproject commit 5d32ba68ece5b2e0f885e305e27231a1deb7bdd5
``` 
4. `git add jslib`
5. `git commit -m "update jslib version"`
6. `git push`

You'll need to do the same for desktop.

Good point about web, I didn't realise we were already using radio buttons there. However, it would be good to do a light refactor of web so that it also uses `passTypeOptions` and `*ngFor`, following the same pattern as browser and desktop. Would you mind submitting a PR for that?

src/app/vault/password-generator.component.html
eliykat, line 37: Ah, I didn't realise you'd done the internationalization here to begin with. Anyway, now that we've got the call to `i18nService.t` in the base class (in jslib), we can remove the `i18n` pipe here. So this line will just be: `{{o.name}}`

eliykat (changes requested):

src/app/vault/password-generator.component.html
knhash: Added back the label.

knhash:
> Good point about web, I didn't realise we were already using radio buttons there. However, it would be good to do a light refactor of web so that it also uses `passTypeOptions` and `*ngFor`, following the same pattern as browser and desktop. Would you mind submitting a PR for that?

Sure thing. I will submit the changes there.

knhash:
@eliykat (WEB-PR)[https://github.com/bitwarden/web/pull/1059] has been raised.

eliykat (approved):
Looks great, thanks! There were some conflicts with jslib (nothing you did), so I took the liberty of sorting those out.

PR #959, opened 2021-06-17
Add password show/hide to reprompt

Hinton:
## Objective
Add a show/hide button to the password field.

### Testing Considerations
We should do a regression sweep of the dialog.

### Screenshots
electron_3LLa9eVPkt

### DIFF
--- a/src/services/electronPlatformUtils.service.ts
+++ b/src/services/electronPlatformUtils.service.ts
@@ -16,11 +16,34 @@ export class ElectronPlatformUtilsService extends BaseElectronPlatformUtilsServi
 
     async showPasswordDialog(title: string, body: string, passwordValidation: (value: string) => Promise<boolean>):
         Promise<boolean> {
+
+        const html = `
+        ${body}
+        <div class="box password-reprompt">
+            <div class="box-content">
+                <div class="box-content-row box-content-row-flex">
+                    <div class="row-main">
+                        <label for="masterPassword">${this.i18nService.t('masterPass')}</label>
+                        <input id="masterPassword" type="password" name="MasterPassword" class="monospaced" required>
+                    </div>
+                    <div class="action-buttons">
+                        <a class="row-btn" href="#" id="toggleVisibility" role="button"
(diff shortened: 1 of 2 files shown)


src/services/electronPlatformUtils.service.ts
```diff
@@ -16,24 +16,67 @@ export class ElectronPlatformUtilsService extends BaseElectronPlatformUtilsServi
...
     async showPasswordDialog(title: string, body: string, passwordValidation: (value: string) => Promise<boolean>):
         Promise<boolean> {
+
+        const html = `
+        ${body}
+        <div class="box password-reprompt">
+            <div class="box-content">
+                <div class="box-content-row box-content-row-flex">
+                    <div class="row-main">
+                        <label for="masterPassword">${this.i18nService.t('masterPass')}</label>
```
kspearrin, line 26: I wonder if XSS is possible via malicious translations here.

cscharf: After a quick review of https://angular.io/guide/security and Snyk's database, it seems Angular 1.8 and below were susceptible to this, but in the way we're using this here I don't think there's a ready opportunity for XSS or template injection. I wondered the same thing, but I _believe_ (not sure) that Angular renders the markup first with the templated values, and then evaluates it as it is added to the DOM, similar to normal HTML (essentially safely sanitizing/encoding the value as expected). @Hinton  may need to weigh in on that.

Hinton: I SweetAlert efficiently bypasses the angular templates and are therefore not protected in the same way. To be safe we should run `Serialize` on the values. Until we've migrate to use Angular templates instead.

Hinton: Follow up: Turns out it's not easy to inject the Angular Sanitizer. We could probably do it by moving the popup logic to `PasswordRepromptService` but I've refactored it to use Modals with Angular templates instead in https://github.com/bitwarden/jslib/pull/417

PR #961, opened 2021-06-17
Redefine cipher "share" to "move to organization"

MGibson1:
# Overview

The term "Share" is confusing to users because it implies maintaining ownership and control of a cipher while allowing others to access it as well. It also implies the ability to revoke access or to stop sharing -- neither of these things is true.

In addition to the share term being misleading, the [`fa-share-alt`](https://fontawesome.com/v4.7/icon/share-alt) icon implies a forking of the cipher item rather than transferring ownership.

When a cipher is "shared" it is irrevocably taken over by an organization. This occurs for good cryptographic and security reasons. The better solution to this is to revise the communication around "sharing." To that end, this work changes the term "share" to "move to organization" and the icon is changed to [`fa-cube`](https://fontawesome.com/v4.7/icon/cube), which is used to indicate a collection. The act of moving an item to an organization is denoted by [`fa-arrow-circle-o-right`](https://fontawesome.com/v4.7/icon/arrow-circle-o-right) rather than `fa-share-alt`.

# Screen shots
#### Organization owned item in list
image

#### Move to organization button
image

#### Move to organization component
image

#### Move to organization success toast
image

### DIFF
--- a/src/app/vault/share.component.html
+++ b/src/app/vault/share.component.html
@@ -1,10 +1,10 @@
-<div class="modal fade" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="shareTitle">
+<div class="modal fade" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="moveSelectedToOrgTitle">
     <div class="modal-dialog" role="document">
         <form class="modal-content" #form (ngSubmit)="submit()" [appApiAction]="formPromise">
             <div class="modal-body">
                 <div class="box">
-                    <div class="box-header" id="shareTitle">
-                        {{'share' | i18n}}
+                    <div class="box-header" id="moveSelectedToOrgTitle">
+                        {{'moveToOrganization' | i18n}}
                     </div>
                     <div class="box-content" *ngIf="!organizations || !organizations.length">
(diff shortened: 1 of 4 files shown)


src/app/vault/share.component.html
```diff
@@ -1,10 +1,10 @@
-<div class="modal fade" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="shareTitle">
+<div class="modal fade" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="moveSelectedToOrgTitle">
     <div class="modal-dialog" role="document">
         <form class="modal-content" #form (ngSubmit)="submit()" [appApiAction]="formPromise">
             <div class="modal-body">
                 <div class="box">
-                    <div class="box-header" id="shareTitle">
-                        {{'share' | i18n}}
+                    <div class="box-header" id="moveSelectedToOrgTitle">
```
eliykat, line 6: This is the `id` you used in the bulk share component on `web`, but this is not a bulk action. I think it should just be `moveToOrgTitle`.

eliykat (changes requested):

eliykat (approved):

PR #967, opened 2021-06-24
fixed the icon for sends being swapped between file and text sends

addisonbeck:
Resolves https://app.asana.com/0/1199950673439194/1200515179495371/f
The icons for Bitwarden Send are backwards on desktop, this flips them around.


cscharf (approved):
Looks good

PR #970, opened 2021-06-29
[macOS] Don't enable secure input when app is not in focus

### LINKED ISSUE #298: macOS Vault Master Password Entry Holds Secure Keyboard Input Open
With the macOS Desktop Vault, version 1.15.2, I'm seeing some new behavior with secure keyboard input or keyboard focus.  I'm on Mojave 10.14.6.

My issue is that Bitwarden keeps keyboard focus active in the Master Password field, even when it's in the background.  When secure keyboard input is active, system-wide keyboard macro software is unable to function (in my case, Keyboard Maestro).

This is a recent break for me, but I'm not sure if secure keyboard input is new, or focus behavior has changed.  As near as I can tell, here's what's happening:

- When Bitwarden loses application focus, it ends the secure input/keyboard focus on the password field

- When Bitwarden regains application focus, it puts keyboard focus back on the field (if it had it when you left; if you click off the master password field to stop keyboard focus, it won't return it when application focus returns)

This is all fine!

But:  It takes a few seconds for Bitwarden to go from application start to fully initialized, where it renders the vault locked UI.  _If it loses application focus during this window_, it still begins with keyboard focus in the master password entry, and thus secure input, even though it's no longer the front application.

The way my applications start up on boot, by the time Bitwarden has fully initialized, some other restarting application has taken focus.  I have to tab in and out of Bitwarden, or manually click off the master password entry text field, before Keyboard Maestro begins working.

I think that adding a check to see if Bitwarden is still the front-most application before putting keyboard focus into the password field will fix this.

eliykat:
## Objective

Fix #298: 
> Bitwarden keeps keyboard focus active in the Master Password field, even when it's in the background. When secure keyboard input is active, system-wide keyboard macro software is unable to function (in my case, Keyboard Maestro).

Secure input is a feature provided by macOS to stop other programs from eavesdropping when entering sensitive information. This is fine when the desktop app is actually in use. However, holding secure input open like this when the app is in the background breaks productivity and accessibility applications, which need to be able to read keypresses to function properly.

Electron seems to handle secure input properly in general. If you're using the app and then click away from it, secure input will be released. This issue only occurs when the app loads or navigates in the background - for example:
* opening the app and then clicking away before Electron loads the page
* when the timeout action occurs

Therefore, this only affects the lock and login pages.

## Code changes

The ideal fix here is to use the Electron API to enable secure input when required, and disable it when it's not required: https://www.electronjs.org/docs/api/app#appsetsecurekeyboardentryenabledenabled-macos

However, this just didn't work for me. Disabling secure input did not have any affect on the behaviour.  Wish I had a better explanation here!

Instead, I've taken the following approach. Secure input is only enabled if a password input element is focused. Therefore, on the lock and login pages, I've delayed setting the default focus on the input elements until the window itself has focus. Once the window has had focus after the initial page load, Electron will properly handle the secure input from there.

Note that we do this for all operating systems, not just macOS. It's only needed for macOS, but it kept it simpler and more consistent just to do it for everyone, and I couldn't see any obvious downsides to this. However I can insert `process.platform === 'darwin'` gates if required.

Related jslib PR: https://github.com/bitwarden/jslib/pull/419. Build failures are caused by jslib changes.

### DIFF
--- a/src/app/accounts/lock.component.html
+++ b/src/app/accounts/lock.component.html
@@ -8,12 +8,12 @@
                     <div class="row-main" *ngIf="pinLock">
                         <label for="pin">{{'pin' | i18n}}</label>
                         <input id="pin" type="{{showPassword ? 'text' : 'password'}}" name="PIN" class="monospaced"
-                            [(ngModel)]="pin" required appAutofocus>
+                            [(ngModel)]="pin" required>
                     </div>
                     <div class="row-main" *ngIf="!pinLock">
                         <label for="masterPassword">{{'masterPass' | i18n}}</label>
                         <input id="masterPassword" type="{{showPassword ? 'text' : 'password'}}" name="MasterPassword"
-                            class="monospaced" [(ngModel)]="masterPassword" required appAutofocus>
+                            class="monospaced" [(ngModel)]="masterPassword" required>
                     </div>
                     <div class="action-buttons">
                         <a class="row-btn" href="#" appStopClick appBlurClick role="button"
--- a/src/app/accounts/lock.component.ts
+++ b/src/app/accounts/lock.component.ts
@@ -33,6 +33,8 @@ const BroadcasterSubscriptionId = 'LockComponent';
     templateUrl: 'lock.component.html',
 })
 export class LockComponent extends BaseLockComponent implements OnDestroy {
+    private deferFocus: boolean = null;
+
     constructor(router: Router, i18nService: I18nService,
         platformUtilsService: PlatformUtilsService, messagingService: MessagingService,
         userService: UserService, cryptoService: CryptoService,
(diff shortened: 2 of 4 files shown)


MGibson1 (commented):
It seems like this issue might impact users of Never lock and biometric as well.

On Mac, your application loses focus when biometric is prompted. there may be a race condition between confirming biometric and refocusing the application. I also think you can change the focus priority order to focus another application after biometric auth.

Finally, never lock will first load the lock page, then the vault itself once the stored key is processed. This provides a way to get all the way to the vault (and the focused search bar) without interacting with the application.

eliykat:
@MGibson1  After our conversation on Slack I tested the following situations:
* Never Lock is enabled. Open app, but click away before page loads. App goes directly to the vault in the background without ever receiving focus.
* Biometric unlock is enabled. Open app, click away before page loads. Biometric auth popup comes up and is unlocked. App goes directly to vault in the background.

In both cases, I confirmed that secure input was not activated, and the bug did not occur. I also tested the search bar directly, and confirmed that because it's not a password input, it doesn't cause any secure input issues.

This was on the production version for convenience. QA can check any regressions.

In general I think we can limit this fix to the lock and unlock pages, as the only pages that:
* are navigated to in the background
* that auto-focus on a password input

MGibson1 (approved):

domparry:
Thank you for this change. I can confirm this solves the secure input issue for me.

Issue #902, opened 2021-05-18
Uninstalling Bitwarden from a user account (non-administrator) in Windows 10 does not remove data in %AppData%\Bitwarden.


SergeantConfused:
## Describe the Bug
When the Bitwarden desktop client is installed in a standard user and then uninstalled from a standard user in Windows 10, the C:\Users\<username>\AppData\Roaming\Bitwarden folder is not emptied or deleted.

## Steps To Reproduce
1. Create a standard user (non-administrator) in Windows 10.
2. Log into that user and download the Bitwarden desktop client installer from the Download page.
3. Install it.
4. Log into your vault.
5. Go to %AppData%\Bitwarden and check if that folder contains any data.
6. Uninstall Bitwarden from Start > Settings > Apps.

## Expected Result
For the %AppData%\Bitwarden folder to be deleted.

## Actual Result
The %AppData%\Bitwarden folder and its contents remain intact.

## Screenshots or Videos
N/A.

## Environment
- Operating system: Windows 10 Pro.
- Installation method: Installed from the .exe installer from here: https://bitwarden.com/download/
- Build Version: 1.26.3.

## Additional Context
When this is performed in an administrator account, the %AppData%\Bitwarden folder is deleted.

baek-sang:
Uninstaller add remove user data check box.

PR #974, opened 2021-06-30
Removing the old setup logic that no longer works

joseph-flinn:
## Summary
With the change to the release process for Desktop (waiting until after a  successful deploy pipeline to publish the release), some of the old logic was still there and making the deploy pipeline fail.

The deploy pipeline will need to be re-examined, but that action item will come out of the retro.


vgrassia (approved):

Nihal247:
Closes #902

vgrassia (approved):

PR #1001, opened 2021-07-23
Use cdk-virtual-scroll for long cipher lists

eliykat:
## Objective

Desktop implementation of https://github.com/bitwarden/browser/pull/1858. **Note:** because the browser PR is already merged, this PR and related PRs need to be actioned before the upcoming release.

Note that Angular-Ivy was already enabled on desktop, so that change does not apply to this PR.

## Code changes
These largely mirror the changes made in browser:
* remove the `infiniteScroll` library and directives
* replace with `cdk-virtual-scroll` library and directives
* update the css for scrollbar theming

In addition to the above, there are the following desktop-specific changes:
* removed `infinite-scroll` directives in `send.component.html`. This looks like some leftover code that should've been removed, because it's not properly wired up. e.g. `loadMore()` does not exist in this component and there is no proper paging logic to make this work. Basically, infinite scroll was not doing anything here. Also, other Angular clients do not use infinite scroll in their Send component.

Note: builds will probably fail due to jslib.

## Related changes
After these changes, infinite scroll will only be used by web, so I've moved all that logic from jslib to the web client. Please also review:
* https://github.com/bitwarden/jslib/pull/436
* https://github.com/bitwarden/web/pull/1094


src/app/vault/ciphers.component.html
```diff
@@ -6,11 +6,9 @@
     </div>
 </div>
 <div class="content">
-    <ng-container *ngIf="(isPaging() ? pagedCiphers : ciphers) as filteredCiphers">
-        <div class="list" *ngIf="filteredCiphers.length" infiniteScroll [infiniteScrollDistance]="1"
-            [infiniteScrollContainer]="'#items .content'" [fromRoot]="true" [infiniteScrollDisabled]="!isPaging()"
-            (scrolled)="loadMore()">
-            <a *ngFor="let c of filteredCiphers" appStopClick (click)="selectCipher(c)"
+    <cdk-virtual-scroll-viewport itemSize="42" minBufferPx="300" maxBufferPx="300" *ngIf="ciphers.length">
```
MGibson1, line 9: shouldn't the buffer size be a multiple of the itemsize?

Also, in my desktop env the calculated item height is 43 before these changes, can we calculate this or is precision not really that important?

MGibson1 (commented):
a couple of questions due to my ignorance

cscharf (dismissed):

src/app/vault/ciphers.component.html
```diff
@@ -30,15 +28,15 @@
...
-            <i class="fa fa-spinner fa-spin fa-3x" *ngIf="!loaded" aria-hidden="true"></i>
-            <ng-container *ngIf="loaded">
-                <i class="fa fa-frown-o fa-4x" aria-hidden="true"></i>
-                <p>{{'noItemsInList' | i18n}}</p>
-                <button (click)="addCipher()" class="btn block primary link">{{'addItem' | i18n}}</button>
-            </ng-container>
-        </div>
-    </ng-container>
+    </cdk-virtual-scroll-viewport>
+    <div class="no-items" *ngIf="!ciphers.length">
```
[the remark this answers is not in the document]
eliykat: The `cdk-virtual-scroll-viewport` is only a container for managing the virtually scrolling list, so there's no reason for this `<div>` to be inside it.

Regarding the old `ng-container` more generally, it did 2 things:
1. `*ngIf="ciphers"` - but that seems unnecessary because this component always instantiates an empty array in the declaration of the `ciphers` property
2. assigned the value of `ciphers`/`pagedCiphers` to `filteredCiphers` - but that is not needed any more because there are no `pagedCiphers`.

That's why I deleted it.

src/app/vault/ciphers.component.html
eliykat: The buffer size doesn't need to be a multiple of the itemSize, it will just render in multiples of the itemSize without exceeding the buffer size. That said, I didn't think through these buffer settings and I'm not sure they're doing anything, so I'll reevaluate here.

The calculated height is 43 for me as well - I had copied `42` from browser without checking. `ItemSize` is used to correctly size the `cdk-virtual-scroll-viewport`, so that the scrollbar matches the expected number of items. Any inaccuracy here could be multiplied over many items and result in the scrollbar being too long (leaving blank space at the end of the list) or too short (cutting off items). I'll look into grabbing the actual height of the first item instead as I don't feel great about hardcoded pixel values.

eliykat:
In addition to my comments above, I've also asked Clayton to have a quick play with this before merging. It doesn't _feel_ quite right to me (maybe slight performance lag?) but I'm not sure if it's in my head or not.

src/app/vault/ciphers.component.html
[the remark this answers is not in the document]
MGibson1: thanks, makes sense

src/app/vault/ciphers.component.html
eliykat: Fixed by the custom scrolling strategy we've discussed

PR #1012, opened 2021-08-04
Add support for setting tray icon in Linux.

gareth-palmer:
The icon used for the tray is the same as the application icon which has a blue background, this does not match other tray icons on most Linux desktop themes which are monochromatic.

Electron does not provide an easy way to set an icon that can be adapted to the current theme in Linux so this allows the user to pick an icon that matches the theme.

Sample light and dark icons have been provided.

Community discussion link: https://community.bitwarden.com/t/add-support-for-changing-the-tray-icon/32207

Depends on https://github.com/bitwarden/jslib/pull/447


Hinton:
Hi,

We are currently in the process of archiving this repository and moving the code and future development to the [bitwarden/clients](https://github.com/bitwarden/clients) repository. This PR will be closed but feel free to re-open it again on [bitwarden/clients](https://github.com/bitwarden/clients).

PR #1026, opened 2021-08-17
Added more suported browsers to the Chrome Webstore link in help menu

### LINKED ISSUE #352: Tab missing in web brower extensions
Brave extension does not appear in the 'Get Web Browser extension" of the Desktop application.

image

djsmith85:
Added Brave, Vivaldi and Tor Browser to the install extension menu in help menu

image

Closes #352

### DIFF
--- a/src/main/menu.main.ts
+++ b/src/main/menu.main.ts
@@ -273,7 +273,7 @@ export class MenuMain extends BaseMenu {
                 label: this.main.i18nService.t('getBrowserExtension'),
                 submenu: [
                     {
-                        label: 'Chrome',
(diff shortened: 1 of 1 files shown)


cscharf:
@djsmith85 , not sure we would want to make this change, as the list is pretty simple and the Chrome Web Store is what supports all of those browsers, so it would make sense to leave it labeled as-is.

djsmith85:
@cscharf  Sure, I just stumbled over the above mentioned issue, while doing some Github house cleaning over the weekend and thought I add it, as I was already fixing the broken link for the Edge store.

PR #1030, opened 2021-08-19
Revert "Autosync Crowdin Translations (#1021)"

joseph-flinn:
## Summary
The Crowdin sync changed the English file which is our source of truth. This reverts commit 026ee2025096c5f109514f440c18147a284d260b.


cscharf (approved):
Wonder if we could have just reverted the `en` file? Others are good because they undo some bad translations in `es`, etc.

PR #1061, opened 2021-09-06
Add issue template and template chooser

Hinton:
## Objective
Simplify the workflow for submitting issues and finding help links by adding a new issue template which utilizes github issue forms.


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -0,0 +1,80 @@
...
+        - Linux
+    validations:
+      required: true
+  - type: input
+    id: os-version
+    attributes:
+      label: Operating System Version
+      description: What version of the operating system(s) are you seeing the problem on?
+  - type: dropdown
+    id: browsers
```
djsmith85, line 60: I don't believe the id is particular important, but might change it to install or install_method

djsmith85 (commented):
Looking very good and I'm happy see the new templates coming. Just a really minor comment

.github/ISSUE_TEMPLATE/bug.yml
Hinton: Nice catch. Yea form what I've seen they aren't important, but they have to be unique.

djsmith85 (approved):

PR #1064, opened 2021-09-08
Use a modal to set the unlock pin

eliykat:
## Objective

Desktop implementation of https://github.com/bitwarden/jslib/pull/477.

## Code changes

* `settings.ts` - move the "set pin" cryptography logic to the base `setPin` component (see jslib PR). Use `modalService` to open the `setPin` component.
* `set-pin.html` - layout for the component
* other misc changes

Build is failing due to jslib changes; jslib needs to be merged first.

## Screenshots

Screen Shot 2021-09-08 at 12 34 29 pm


eliykat:
@Hinton Using the `appAutoFocus` directive in the modal is giving me the following error: 

Screen Shot 2021-09-08 at 1 23 23 pm

I followed the same pattern as your password reprompt modals, do you know why I'm getting this error now?

src/app/components/set-pin.component.html
```diff
@@ -0,0 +1,44 @@
+<div class="modal fade" tabindex="-1" role="dialog" aria-modal="true" aria-labelledby="confirmUserTitle">
```
Hinton, line 1: The `aria-labelledby` seems incorrect. I don't believe we have any title for this modal.

Hinton (changes requested):

Hinton:
@eliykat Hmm, not sure what causes the `ExpressionChanged` error. I know we call `detectChanges` in `dynamic-modal.component` but that doesn't seem connected from my quick testing.

Wrapping the focus part of the autofocus directive with a `setTimeout` seems to work though but that's only a work-around. I debugged it slightly and it seems when setting focus, angular marks the input as touched which causes the issue.

Interestingly it only occurs when we have multiple modals open.

eliykat:
@Hinton I've found the "right" way to do this, which should also fix some accessibility issues on dynamic modals.

Currently tab focus is not trapped within dynamic modals - you can tab out of them to other elements in the background of the page. I've created a ticket here: https://app.asana.com/0/1169444489336079/1200947365681007/f

We can use the `cdkTrapFocus cdkTrapFocusAutoCapture` directives to trap the tab focus within an element. They're part of `angular/cdk/a11y`, which we already use ([docs reference](https://material.angular.io/cdk/a11y/api#CdkTrapFocus)). And if we're using those, we can then use `cdkFocusInitial` to set the initial focus on a child component, without getting any change detection errors.

I've updated the `settings.component` and `set-pin.component` just so that these PRs can be considered complete and ready for merging. However, I think it would be preferable to apply these 2 directives at the level of the `dynamic-modal.component` so that they automatically apply to all modals without having to remember to put them in every modal template. (On the other hand, there's already a lot of duplication in template structure, so you might disagree here.)

Hinton (approved):
I think we could move trap focus to dynamic component, but would need to be tested to verify it doesn't break anything.

PR #1084, opened 2021-09-17
Use explicit import paths

Hinton:
## Objective
Follow up to https://github.com/bitwarden/jslib/pull/490 which removed the `index.ts` files which means we need to use the explicit import paths.


eliykat (approved):

eliykat (changes requested):
Please fix linting

eliykat (approved):

PR #1089, opened 2021-09-24
Use webfonts from jslib instead of downloading them using gulp

Hinton:
## Objective
`gulp-google-webfonts` broke recently due do a change in the google webfonts api response. Rather than trying to patch the gulp plugin we should just remove it altogether. Since in several of our projects it's the only use we had for gulp.

Depends on: https://github.com/bitwarden/jslib/pull/500


addisonbeck (approved):

eliykat:
@Hinton I'm hitting merge on this one just so I can unblock some desktop work.

PR #1169, opened 2021-11-19
Hide password input when using key connector

Hinton:
## Type of change
- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Hides the password input when using key connector.

Resolves https://app.asana.com/0/1201363553288890/1201394615545878
Depends on https://github.com/bitwarden/jslib/pull/560

## Screenshots

image

## Before you submit

### DIFF
--- a/src/app/accounts/lock.component.html
+++ b/src/app/accounts/lock.component.html
@@ -1,10 +1,10 @@
 <form id="lock-page" (ngSubmit)="submit()">
     <div class="content">
         <p aria-hidden="true"><i class="fa fa-lock fa-4x text-muted"></i></p>
-        <p>{{(pinLock ? 'yourVaultIsLockedPinCode' : 'yourVaultIsLocked') | i18n}}</p>
+        <p>{{'yourVaultIsLocked' | i18n}}</p>
         <div class="box last">
             <div class="box-content">
-                <div class="box-content-row box-content-row-flex" appBoxRow>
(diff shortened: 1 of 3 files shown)


src/app/accounts/lock.component.html
```diff
@@ -1,10 +1,10 @@
 <form id="lock-page" (ngSubmit)="submit()">
     <div class="content">
         <p aria-hidden="true"><i class="fa fa-lock fa-4x text-muted"></i></p>
-        <p>{{(pinLock ? 'yourVaultIsLockedPinCode' : 'yourVaultIsLocked') | i18n}}</p>
+        <p>{{'yourVaultIsLocked' | i18n}}</p>
```
djsmith85, line 4: I think we should still show a different text when unlocking with PIN as it has no effect in regards to KeyConnector

djsmith85 (changes requested):
If a user has biometrics enabled, they most likely would want to use it to log in. So if we move the biometrics button to the top, then we would only need to hide the submit button. This would greatly simplify the code and improve UX.

src/app/accounts/lock.component.html
djsmith85: @Hinton Spoke to Danielle and she is fine with only having one label. Do you think we could simplify the structure for the biometrics button with regards to my comment above? If not, it's looking good for approval

Hinton: I think we could do that. Lets loop it by design.

djsmith85 (approved):

PR #1182, opened 2021-12-07
Add autotype basic support

luisfontes19:
## Type of change
- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective
Add basic autotype support, for macos for now

Please check [JSLib PR 570](https://github.com/bitwarden/jslib/pull/570) as its needed as well

## Code changes
* native extension (addon), for now only for MacOS, to do the typing using low level OS's apis.
* added the extension build step at the package.json scripts (this may need testing when bundling the final electron app)
  * To achieve this, I had to disable the CleanWebpackPlugin, as both the electron-rebuild and webpack were cleaning the build folder. Instead, now a manual clean is done, before the "build" script. This may need to be tweaked.
* For now as an initial version the autotype is not configurable, this means that it will switch window (cmd+tab), type username, hit tab, type password, hit enter. There isn't a delay between keystrokes for now, so its typing insanely fast
* There's a button in the item toolbar (src/app/vault/view.component.html) with a keyboard icon to do the autotype. It was created a method to check if the autotype is available (as for now only MacOS is implemented). The button will be hidden if not supported

## Screenshots
image

## Testing requirements
Testing the build and bundle process (specially in other platforms to make sure nothing is broken)

## Before you submit
- [x] This change requires a **documentation update** (notify the documentation team)
- [x] This change has particular **deployment requirements** (notify the DevOps team)


djsmith85:
Hi @luisfontes19 and thank you for your contribution. 

Per our [Contribution Guidelines](https://github.com/bitwarden/desktop/blob/master/CONTRIBUTING.md) we usually ask for a post in the Github Contributions category of our forums, to describe the proposed changes, so the community and Bitwarden team members can discuss this upfront before writing any code. 

Could you please describe your feature in more detail, either here or as mentioned before in the [Community forums](https://community.bitwarden.com/c/github-contributions/7).

Kind regards,
Daniel

luisfontes19:
Hi @djsmith85, thanks for the notice

If it's ok then i'll do it here, for more visibility then :) 
I started the PR for myself as I wanted to try bitwarden, but autotype is a deal breaker for me. I use keepassxc with autotype everywhere. 

The idea is to have an autotype feature like in keepass. This way you don't need a browser extension, and it will work on all windows, not just the browser (with the extension). 


I created a button on the entries toolbar, to trigger the autotype. Behind the seens it uses ipcRenderer to invoke a method in the main electron's thread which in turn will call a native extension.
The native extension is a node addon in c++ to invoke the OS api's. This needs be replicated for other OS's. For now I only did it for macOS.

The autotype uses low level OS API's to simulate keyboard keystrokes to fill the username and password fields. As this is just a PoC, you can't configure the sequence like in keepass yet. When triggered it will switch windows with cmd+tab, type the username, hit tab, type the password and press enter. With time, and if this gets approved we can work on a window to configure this sequence (keepass compatible as well), implement on other OS's.

Also, for anybody testing it:
I spent a huge time trying to figure out some issues with the native extension. In case you need to test it, make sure you have the same node and npm version as the ones used by electron otherwise the native extension will not work.
At the time of this PR the required node version to make it work with electron is '14.17.0' which is the same as defined in .nvmrc

luisfontes19:
preview

DOWIT-JoelFrojmowicz:
Please, add autotype to Windows APP. It's a must have feature.

Hinton:
Hi @luisfontes19,

Thanks for expressing interest in this feature, and for implementing a proof of concept for macOS.

This feature uses alt+tab, making it to fragile for us to support long term as it requires a specific order of the windows. We are also not looking into adding C++ code to our codebase.

As of right now I cannot give any further insight into when we plan on releasing this feature, as we are still performing research into how to implement this in a maintainable, secure and reliable way.

Thanks again for showing interest and opening a PR, but this PR will be closed for now, as we are in the process of re-organizing our GitHub repositories.

PR #1221, opened 2022-01-06
Remove webpack-node-externals and use asarUnpack

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Changes how we build and bundle `node_modules` with the main processes. Instead of packaging `node_modules` using `webpack-node-externals` we let webpack bundle them into the `main.js`. This allows us to get rid of dependencies in the second `package.json`, and also removes the `node_modules` directory from `app.asar` which had 1500 files and was 18mb large.

Additionally I also added `asarUnpack` to unpack all our native node_modules, which should resolve some issues in certain environments where binaries files placed in `tmp` aren't treated well.

## Testing requirements

I've tested this with nsis install on Windows, and using DMG and masdev on macOS, and deb on linux. But we will have to verify the other environments as well.

## Before you submit


djsmith85 (dismissed):
LGTM

Hinton:
`npm run electron` is failing due to `electron-rebuild`. We should look into using `webpack dev server` instead which will solve this, and be more in-line with how other electron projects handles this.

djsmith85 (approved):
Changes look good. Tested this with the output from `pack:win:ci` and also executing `npm run electron`. Both work fine.

PR #1231, opened 2022-01-12
Fix logging out of all accounts instead of one

djsmith85:
## Type of change

- [X] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
When a user clicks on "Lock all vaults" they are brought onto the lock screen of the last active account. Then clicking on Logout would log out all users instead of switching to the other locked account.

Also when logged in with two users and then logging out via the context menu it would log you out of all accounts.

Asana tasks: 

https://app.asana.com/0/1201648796371593/1201627172404375/f
https://app.asana.com/0/1201648796371593/1201642542946835/f

## Code changes
- **src/app/app.component.ts:** When logOut is triggered and the userId is not provided. Retrieve currentUserId

## Testing requirements
Have multiple accounts logged in.

- **1st case**
Log out via the context menu and it should switch to the other account.

- **2nd case**
Lock all vaults and you get brought back onto the lock screen, clicking logout there should also switch to the other accounts lock screen.

- **3rd case**
In no scenerio should the email address on the lock screen look like `__$1__`
## Before you submit

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -434,6 +434,10 @@ export class AppComponent implements OnInit {
   }
 
   private async logOut(expired: boolean, userId?: string) {
+    if (!userId) {
+      userId = await this.stateService.getUserId();
+    }
+
     await Promise.all([
       this.eventService.uploadEvents(userId),
       this.syncService.setLastSync(new Date(0), userId),


addisonbeck:
I'm not sure what setting the userId here is really changing. If a null userId is passed in to any stateService call it is reconciled as the activeUserId already.

Also: I'm not able to recreate this bug in master. Is it possible this was fixed already elsewhere?

addisonbeck:
I'm not able to recreate the logout from lock screen bug either. I would imagine both of these were fixed with the change to clear access token on state clean that I added a few days ago.

addisonbeck (approved):
Daniel and I spoke about this on Slack and decided to move forward with this change

PR #1241, opened 2022-01-17
[Account switching] Hide active account

djsmith85:
## Type of change

- [ ] Bug fix
- [X] New feature development
- [ ] Other

## Objective
The current active account should only be visible on the switcher and not show up as an item in the dropdown

Asana task: https://app.asana.com/0/1201648796371593/1201665881786813/f

## Code changes

- **src/app/layout/account-switcher.component.html:** Tie the visibility of the dropdown to the visibility of the account switcher above it.
- **src/app/layout/account-switcher.component.ts:** 
  - Only show the switcher when an active account has been set (at least one account logged in)
  - Remove the current active account from the dropdown
  - No longer need to check if selected account is active before switching

## Screenshots
- **BEFORE:**
image

- **AFTER:**
image

## Testing requirements
The currently active account should only appear on the account switcher but never in the dropdown. As soon as I have switched to another account, the account that was active before should be shown in the dropdown.

The switcher usually disappeared when no active account was present. The dropdown should now also close iof it was open before

## Before you submit

### DIFF
--- a/src/app/layout/account-switcher.component.ts
+++ b/src/app/layout/account-switcher.component.ts
@@ -57,11 +57,7 @@ export class AccountSwitcherComponent implements OnInit {
   serverUrl: string;
 
   get showSwitcher() {
-    if (!Utils.isNullOrWhitespace(this.activeAccountEmail)) {
-      return true;
-    }
-
-    return false;
+    return !Utils.isNullOrWhitespace(this.activeAccountEmail);
   }
 
   constructor(
@@ -104,11 +100,7 @@ export class AccountSwitcherComponent implements OnInit {
   }): Promise<{ [userId: string]: SwitcherAccount }> {
     const switcherAccounts: { [userId: string]: SwitcherAccount } = {};
     for (const userId in baseAccounts) {
-      if (userId == null) {
-        continue;
-      }
-
-      if (userId === (await this.stateService.getUserId())) {
+      if (userId == null || userId === (await this.stateService.getUserId())) {
         continue;
       }
 
--- a/src/app/layout/account-switcher.component.ts
+++ b/src/app/layout/account-switcher.component.ts
@@ -96,9 +96,6 @@ export class AccountSwitcherComponent implements OnInit {
   async switch(userId: string) {
     this.toggle();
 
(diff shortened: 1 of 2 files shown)


addisonbeck:
@djsmith85 There is an issue here with the positioning of the account dropdown when there is only one authenticated account. It appears offset to the left too much, and disconnected from the toggle icon completely (see screenshot). I see two paths forward for this:

1. Some hack or upgrade to the min-width of the dropdown so it always covers enough distance to appear connected.
2. Moving the cdkOverlay init logic out of the template and into the component, and making use of the position field exposed by angular. This may prove challeneging because the [docs](https://v7.material.angular.io/cdk/overlay/overview) are pretty bad. I was trying to piece something together last week by just reading the [code](https://github.com/angular/components/blob/master/src/cdk/overlay/position/connected-position.ts) but didn't make it anywhere.

Screen Shot 2022-01-18 at 8 48 06 AM

src/app/layout/account-switcher.component.ts
```diff
@@ -105,6 +107,11 @@ export class AccountSwitcherComponent implements OnInit {
       if (userId == null) {
         continue;
       }
+
+      if (userId === (await this.stateService.getUserId())) {
+        continue;
+      }
+
```
addisonbeck, line 114: I recommend merging this conditional with the one above it: `(userId == null || userId == activeUserId)`

src/app/layout/account-switcher.component.ts
```diff
@@ -56,7 +57,11 @@ export class AccountSwitcherComponent implements OnInit {
   serverUrl: string;
 
   get showSwitcher() {
-    return this.accounts != null && Object.keys(this.accounts).length > 0;
+    if (!Utils.isNullOrWhitespace(this.activeAccountEmail)) {
+      return true;
+    }
+
+    return false;
```
addisonbeck, line 64:
```suggestion
    return !Utils.isNullOrWhitespace(this.activeAccountEmail));
```

djsmith85:
@addisonbeck As discussed I've create a new Asana task to tackle the (min)-width issue.

addisonbeck (approved):

PR #1258, opened 2022-01-22
Unset active account when adding additional accounts

addisonbeck:
## Type of change
- [ ] Bug fix
- [ ] New feature development
- [x] Other

## Objective
Currently we maintain an active account when adding additional accounts via account switching.
Instead, we would like to unset the active account when adding accounts.

## Code changes
* Kinda unrelated, but I noticed the "Max account limit reached" warning was a link, which gave it a hover effect I don't think we want. I changed this to a span and added the css needed to keep it looking the same as before, just without the hover.
* Unset the active account when clicking "Add account" and redirecting to login.
* Show a "Switch Account" helper text as the account switcher header if no active account is present.
* Hide the "Add account" button or "Max account limit reached" warning if already on the login screen adding an additional account.

## Screenshots
https://user-images.githubusercontent.com/15897251/150654210-6e520fc4-5427-4af0-b2b9-6b892debe657.mov

## Before you submit

### DIFF
--- a/src/app/layout/account-switcher.component.html
+++ b/src/app/layout/account-switcher.component.html
@@ -5,15 +5,20 @@
   #trigger="cdkOverlayOrigin"
   [hidden]="!showSwitcher"
 >
-  <app-avatar
-    [data]="activeAccountEmail"
-    size="25"
-    [circle]="true"
-    [fontSize]="14"
-    [dynamic]="true"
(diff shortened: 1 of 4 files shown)


djsmith85 (approved):
Looks great! Love the calm and slighty eerie music 😉

PR #1266, opened 2022-01-25
Add support for handling multiple accounts with native messaging

Hinton:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Adds support for sending the correct encryption key depending on which account requests it. (this allows you to have selected into one account, but use biometrics from the browser with another)

Depends on https://github.com/bitwarden/jslib/pull/633

## Before you submit

- [ ] I have checked for **linting** errors (`npm run lint`) (required)


src/services/nativeMessaging.service.ts
```diff
@@ -41,16 +55,17 @@ export class NativeMessagingService {
...
     const rawMessage = msg.message;
 
     // Request to setup secure encryption
-    if (rawMessage.command === "setupEncryption") {
+    if ("command" in rawMessage && rawMessage.command === "setupEncryption") {
       const remotePublicKey = Utils.fromB64ToArray(rawMessage.publicKey).buffer;
 
       // Valudate the UserId to ensure we are logged into the same account.
-      if (rawMessage.userId !== (await this.stateService.getUserId())) {
+      const userIds = Object.keys(await firstValueFrom(this.stateService.accounts));
```
addisonbeck, line 67: You just need to get the current set of accounts, right? If so, this could probably just be `this.stateService.accounts.getValue()` and cut out the need to import rxjs

addisonbeck (approved):

PR #1268, opened 2022-01-25
Adjust an iterator used to check for locked vaults on reloadProcess

addisonbeck:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Simply iterator adjustment to use a key instead of an index


Hinton (approved):
🏆

PR #1273, opened 2022-01-27
[bug] Patch the windows menu bar regressions

addisonbeck:
https://app.asana.com/0/1201648796371593/1201638560333659

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Several regressions with the windows menu bar need to be cleaned up from the menu bar rewrite.

## Code changes
* Combine shared items between the Bitwarden menu on mac and the File menu everywhere else into a FirstMenu class that both extend
* Use the more generic isMac instead of isMacAppStore for some controls
* Add FirstMenu items to FileMenu if !isMac()
* Flip legal to only show for the Mac App Store instead of for everything else
* Don't show the Bitwarden menu unless we are on Mac

## Screenshots
No screenshots, I was depending on help to get this tested 😔

## Testing requirements
We should do a regression on all menus for all clients/builds and make sure everything is as expected

## Before you submit


src/main/menu.file.ts
```diff
@@ -1,39 +1,60 @@
...
-  private readonly _isLocked: boolean;
+    if (!isMac()) {
+      items = [
+        ...items,
+        ...[
+          this.separator,
+          this.settings,
+          this.lock,
+          this.lockAll,
+          this.logOut,
```
Hinton, line 39: This belongs to Account.

src/main/menu.file.ts
```diff
@@ -1,39 +1,60 @@
...
-  private readonly _messagingService: MessagingService;
-  private readonly _isLocked: boolean;
+    if (!isMac()) {
+      items = [
+        ...items,
+        ...[
+          this.separator,
+          this.settings,
+          this.lock,
+          this.lockAll,
```
Hinton, line 38: I don't really like that we have lock in `file` to begin with. It feels like it should belong to Account especially if we add more of them.

src/main/menu.bitwarden.ts
```diff
@@ -24,6 +25,7 @@ export class BitwardenMenu implements IMenubarMenu {
     items.push(this.lock);
     items.push(this.lockAll);
     items.push(this.logOut);
```
Hinton, line 27: logout doesn't belong here. It's under account.

addisonbeck:
Each of these changes was a requirement for the account switching work, I believe because the controls now effect multiple accounts and the Account menu should only target the active account. I don't think this is a good change to try and get everyone on board with and make the day we are supposed to cut RC branches.

Hinton (approved):

Hinton:
Is there potentially a conflict if a menu is not visible, but has a specific role?

addisonbeck:
I didn't observe this behavior, it seems like visible is more than just a [hidden] and leaves the item out of the menu completely.

PR #1278, opened 2022-02-01
[bug] Inactive accounts with power-based timeout settings are not timing out

addisonbeck:
https://app.asana.com/0/1201648796371593/1201747143977580

This will need to be cherry picked into `rc`

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
`power.main` currently only checks the active account. It needs to check all authenticated accounts.

This is a bit tricky, because the main process's `stateService` instance isn't the same one used by the renderer, and the renderer is the one that maintains in memory account state. It works with currently for the active account because all information needed is pulled from disk storage, but the authenticated accounts list lives in memory. 

To overcome this, I decided to move the bulk of power.mains's timeout behavior into app.component and communicate via messages. This lets us use the same stateService everything else is using, that has the accounts list.

## Code changes
* Relocate some logic from power.main into messages handled by the app component

## Before you submit

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -339,6 +339,45 @@ export class AppComponent implements OnInit {
               this.router.navigate(["vault"]);
             }
             break;
+          case "systemSuspended":
+            for (const userId in this.stateService.accounts.getValue()) {
+              if (userId == null) {
+                continue;
+              }
+              const options = await this.getVaultTimeoutOptions(userId);
+              if (options[0] === -3) {
+                options[1] === "logOut"
+                  ? this.messagingService.send("logout", { expired: false, userId: userId })
+                  : this.messagingService.send("lockVault", { userId: userId });
+              }
+            }
+            break;
+          case "systemLocked":
+            for (const userId in this.stateService.accounts.getValue()) {
+              if (userId == null) {
+                continue;
+              }
+              const options = await this.getVaultTimeoutOptions(userId);
+              if (options[0] === -2) {
+                options[1] === "logOut"
+                  ? this.messagingService.send("logout", { expired: false, userId: userId })
+                  : this.messagingService.send("lockVault", { userId: userId });
+              }
+            }
+            break;
+          case "systemIdel":
+            for (const userId in this.stateService.accounts.getValue()) {
+              if (userId == null) {
+                continue;
+              }
+              const options = await this.getVaultTimeoutOptions(userId);
+              if (options[0] === -4) {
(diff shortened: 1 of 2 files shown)


src/app/app.component.ts
```diff
@@ -339,6 +339,45 @@ export class AppComponent implements OnInit {
...
+              if (userId == null) {
+                continue;
+              }
+              const options = await this.getVaultTimeoutOptions(userId);
+              if (options[0] === -3) {
+                options[1] === "logOut"
+                  ? this.messagingService.send("logout", { expired: false, userId: userId })
+                  : this.messagingService.send("lockVault", { userId: userId });
+              }
+            }
```
djsmith85, line 353: This logic is repeated for the systemCommands, with the only difference being the value of options[0] checked. This could be put into a separate method with a param to equal options[0]. 

Going further instead of sending a message to itself, the logic inside of the `logout` and `lockVault` commands could also be put into methods. These could then be called from the `cases` and from the new method proposed above. 
This would obviously only be an option if no other part of the code relies on receiving `logout` and `lockVault` messages.

djsmith85 (changes requested):
Great idea to put the powerMonitor handling into the app.components. I hadn't thought about the 2 stateService instances yesterday.

src/app/app.component.ts
```diff
@@ -339,6 +339,45 @@ export class AppComponent implements OnInit {
...
+              }
+              const options = await this.getVaultTimeoutOptions(userId);
+              if (options[0] === -2) {
+                options[1] === "logOut"
+                  ? this.messagingService.send("logout", { expired: false, userId: userId })
+                  : this.messagingService.send("lockVault", { userId: userId });
+              }
+            }
+            break;
+          case "systemIdel":
```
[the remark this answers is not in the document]
addisonbeck: Good catch

src/enums/vaultTimeoutPreset.ts
```diff
@@ -0,0 +1,15 @@
+export enum VaultTimeoutPreset {
+  OneMinute = 1,
+  FiveMinutes = 5,
+  FifteenMinutes = 15,
+  ThirtyMinutes = 30,
+  OneHour = 60,
+  FourHours = 340,
```
djsmith85, line 7: Should be 240 😉

djsmith85 (changes requested):

djsmith85 (approved):
Great work!

PR #1280, opened 2022-02-01
[bug] Attempt to resolve windows portable build issues

addisonbeck:
https://app.asana.com/0/1201648796371593/1201711275724167

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Vault timeouts aren't working for some builds, namely windows portable. It seems to work fine building from master, and on other clients through release pipelines like Testflight, so this one is a bit of an enigma.

A data.json file shared from QA showed no lastActive value for any accounts, which would stop vaults from timing out - making this fix more a question of "why isn't lastActive being saved?"

## Code changes
I _think_ it's because we aren't always starting up our activity checks, and instead block them behind a not null activeUserId. I'm not sure why this wouldn't break on all clients, but ¯\\_(ツ)_/¯

This change sets the activeUserId value used in app.component to a class level variable, unblocks the recordActivity calls in ngZone, and instead ensures we have an active user in the recordActivity function itself.

## Testing requirements
I tested this locally to make sure timeouts were still working, but we won't know the bug is resolved until we can test using artifacts from a build.

## Before you submit

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -96,6 +96,7 @@ export class AppComponent implements OnInit {
   private modal: ModalRef = null;
   private idleTimer: number = null;
   private isIdle = false;
+  private activeUserId: string = null;
 
   constructor(
     private broadcasterService: BroadcasterService,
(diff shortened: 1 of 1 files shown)


addisonbeck:
AJ tested the artifact from this workflow and confirmed this fixed the lock issue

djsmith85 (approved):
Code looks good. I can also confirm that with the build artifact the locking works.
Tested with 2 accounts, one set to 1 minute lock and the other to 5 minutes.

Issue #1325, opened 2022-02-11
Desktop app's new tab focus order creates usability issues


aiuax:
### Steps To Reproduce

1. Open Bitwarden desktop app for Windows (v1.31.0)
2. Enter master password
3. Once unlocked, cursor is appropriately placed on 'Search Vault' text entry field.
4. Press `TAB` three times
5. Note that element focus does not move to the item selection panel but simply moves down the category/folder list

### Expected Result

Previously, tab order was set such that pressing tab three times would navigate to the item selection panel, allowing quick selection of items for automatic entry.

In 1.31.0, this is now impossible to achieve and makes keyboard navigation infeasible.

### Actual Result

Attempting to use the `TAB` key to navigate to a searched entry is unwieldy, as it now requires dozens of presses to navigate to the item selection panel.  Pressing `TAB` several times when the search text entry field is focused instead navigates down the left (category/folder) column.

### Additional Context

This is a reversion in terms of user experience.  Keyboard navigation has become less friendly and interacting with individual entries has become extremely difficult.

### Operating System

Windows

### Installation method

Direct Download (from bitwarden.com)

### Build Version

1.31.0

keywordnew:
+1 to this. My experiencing a variation of the original issue, but these are essentially the same. I account for the difference below with only those sections that are different.

### Steps To Reproduce

1. Navigate to search field.
2. Make search to filter entry list
3. Press `tab` once
4. Focus moves to the new user selection dropdown
5. Press `tab` once
6. Focus moves to the sidebar
7. Press `tab` again n times
6. Focus moves down the sidebar

### Expected Result

1. Navigate to search field.
2. Make search to filter entry list
3. Press `tab` once
4. Focus changes to the first entry of the filtered entry list
5. Press `tab` n times
6. Focus moves down the filtered entry list

### Operating System

MacOS

### Build Version

1.31.2

allparadigm:
Confirm affects Flatpak on linux as well as Windows

No longer any efficient way to navigate from keyboard

Hinton:
Hi,

Thanks for providing this feedback, we are discussing internally what the preferred tab order should be.

addisonbeck:
Thanks for the report, y'all. We're on this and will have a patch release with a fix out ASAP.

keywordnew:
With a predictable double tap on `tab`, this is still hugely preferable to the current state reported in this issue, thank you for merging a fix.

[...]

PR #1340, opened 2022-02-14
[bug] Fix tab order regression

addisonbeck:
Resolves https://github.com/bitwarden/desktop/issues/1325
https://app.asana.com/0/1201804463288267/1201808704441945

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Moving the search bar outside the context of the items list has causes a regression of our tabbing behavior. 

Previously, tabbing operated as such:
1. Search bar
2. Item list
3. Add item button
4. Item details (if applicable)
5. Groupings

Now, tabbing operates as:
1. Search bar
2. Account switcher
3. Groupings
4. Items
5. Add item button
6. Item details (if applicable)

This is certainly a worse UX. This PR puts things back the way they were, with the exception that the account switcher still comes before the item list in the focus order.

## Code changes
Ultimately, keyboard behavior in the desktop client is and was already poor - and this issue sparked a design conversation about expanding and improving that use case. This is something we're going to be taking a look at later and improving across the board, but for now has just been returned to the way it was in the last release. 

To achieve this, instead of putting tabIndex on everything in the entire application I moved the groupings component to the bottom of its template and added an explicit item order to the flex rules for the page. This lets us put groupings at the bottom of the DOM, making it the last thing to get tab focus, but still the first left-aligned item on screen.

I also updated jslib just because

## Before you submit

### DIFF
--- a/src/app/vault/vault.component.html
+++ b/src/app/vault/vault.component.html
@@ -1,17 +1,4 @@
 <div id="vault" class="vault" attr.aria-hidden="{{ showingModal }}">
-  <app-vault-groupings
-    id="groupings"
-    class="groupings"
-    (onAllClicked)="clearGroupingFilters()"
-    (onFavoritesClicked)="filterFavorites()"
-    (onCipherTypeClicked)="filterCipherType($event)"
-    (onFolderClicked)="filterFolder($event.id)"
-    (onAddFolder)="addFolder()"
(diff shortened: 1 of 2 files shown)


Hinton (approved):
🤷css magic.

PR #1369, opened 2022-02-22
Add eslint

Hinton:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Replace tslint with eslint. I went with the default eslint rules and added some of the old rules as well. I didn't add `member-ordering`, I'm open to adding it, but we disable it in a few places already since js needs to be organized a specific way, so unsure how much value it would provide.

I mainly preferred to ignore eslint errors vs fixing them. We should remove any comments when we modify nearby code but it seemed a waste of time to test everything now.

Biggest change to before is that imports are now automatically sorted and grouped using eslint. This results in a slightly changed grouping than we are used to. I think the benefits of automating it is an improvement.

## Before you submit

- [ ] I have checked for **linting** errors (`npm run lint`) (required)
- [ ] I have added **unit tests** where it makes sense to do so (encouraged but not required)


src/app/app.component.ts
```diff
@@ -341,12 +341,16 @@ export class AppComponent implements OnInit {
               this.router.navigate(["vault"]);
             }
             break;
+          }
           case "systemSuspended":
             await this.checkForSystemTimeout(systemTimeoutOptions.onSuspend);
+            break;
```
Hinton, line 347: @addisonbeck I'm assuming these were accidentally omitted?

addisonbeck: This was not intentional, we definitely want break statements here.

jlf0dev (approved):

PR #1379, opened 2022-02-25
[BEEEP] Add native rust module

Hinton:
## Objective
As we continue to expand the Desktop application we run into scenarios where available npm modules and electron lack functionality we would want to use. The step forward would be to write our own code which can either be done natively with c++ or rust. Or by using any of the node-ffi libraries. I've found it easier to most often write native code than struggling with ffi.

This PR introduces a native module called `desktop_native` which is written in Rust. It uses Napi-rs to create a native node.js module. Which allows us to fairly seamlessly jump between rust and node.

It currently provides a drop-in replacement of the `keytar` module rewritten in rust.

Depends on https://github.com/bitwarden/jslib/pull/706

EC-141

### Code Changes
- Moved menu files to subdirectory.
- Moved biometric code to desktop in order to ensure compatability with the new module.

### Testing Considerations
We'll want to do a full regression sweep around storing, retrieving and deleting items from the secure storage.


desktop_native/src/biometric/mac.rs
```diff
@@ -0,0 +1,20 @@
+pub async fn available() -> bool {
+    false
+}
+
+pub async fn verify() -> bool {
```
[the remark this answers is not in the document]
MGibson1: We can keep as module, but we don't get the same interface guarantees

workaround thread: https://internals.rust-lang.org/t/module-traits/7532/3

desktop_native/src/biometric/win.rs
```diff
@@ -0,0 +1,66 @@
...
+            UserConsentVerifierAvailability::DeviceBusy => true,
+            _ => false,
+        },
+    }
+}
+
+pub async fn verify(
+    message: &str,
+    window_handle: isize,
+) -> std::result::Result<bool, Box<dyn std::error::Error + Send + Sync>> {
```
MGibson1, line 27: I think `anyhow::Result` is much friendlier than this https://docs.rs/crate/anyhow/1.0.26

desktop_native/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[target.'cfg(windows)'.dependencies.windows]
+version = "0.32.0"
+features = [
+    "alloc",
+    "Foundation",
+    "Security_Credentials_UI",
+    "Win32_Foundation",
+    "Win32_System_WinRT",
+]
```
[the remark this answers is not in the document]
Hinton: You can't have newlines in toml, so that would be a fairly long line.
```toml
neon = { version = "0.10.0-alpha.3", default-features = false, features = ["napi-6", "promise-api", "task-api"] }
```

Only reason to have it in a separate table.

desktop_native/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+After building desktop_native, you can explore its exports at the Node REPL:
+
+```sh
+$ npm install
+$ node
+> require('.').hello()
+"hello node"
+```
+
```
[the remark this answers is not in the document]
Hinton: We should rewrite the whole readme, this is just the scaffolding from neon.

desktop_native/src/biometric/mac.rs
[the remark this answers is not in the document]
Hinton: Yea ... I'm not certain the best way of handling it. I think we need to implement it for both mac and windows to get a feel for the differences. Worst case we'll have two methods, `windowsVerify` and `macVerify`.

desktop_native/src/biometric/mac.rs
```diff
@@ -0,0 +1,20 @@
...
+    #[tokio::test]
+    async fn available() {
+        assert_eq!(false, super::available().await)
+    }
+
+    #[tokio::test]
+    async fn verify() {
+        assert_eq!(false, super::verify().await)
+    }
+}
```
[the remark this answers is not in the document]
Hinton: Won't work if it has the same name right? They would be circular.

desktop_native/src/biometric/win.rs
Hinton: Yes most likely. I wanted to avoid adding further dependencies while hacking it togheter. Certanly something we should look into cleaning up.

desktop_native/Cargo.toml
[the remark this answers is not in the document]
MGibson1: You can, you just need to split it where it accepts them

```toml
neon = { version = "0.10.0-alpha.3", default-features = false,features = [
  "napi-6",
  "promise-api",
  "task-api",
] }
```

although I don't really like that syntax

desktop_native/src/biometric/mac.rs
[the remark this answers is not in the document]
MGibson1: I'm not recommending that magic, but some means of forcing the methods to accept the same arguments

desktop_native/src/biometric/mod.rs
```diff
@@ -0,0 +1,24 @@
...
+use unix as imp;
+#[cfg(windows)]
+use win as imp;
+
+#[cfg(any(target_os = "redox", unix))]
+mod unix;
+#[cfg(not(any(target_os = "redox", unix, windows)))]
+mod unknown;
+#[cfg(windows)]
+mod win;
```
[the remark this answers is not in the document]
Hinton: Found a nicer syntax for this.

desktop_native/src/password/macos.rs
```diff
@@ -0,0 +1,43 @@
+use anyhow::Result;
+use security_framework::passwords::{
+    delete_generic_password, get_generic_password, set_generic_password,
+};
+
+pub async fn get_password<'a>(service: &str, account: &str) -> Result<String> {
```
MGibson1, line 6: I think all these methods are blocking, does it do anything to make these methods async?

desktop_native/src/password/unix.rs
```diff
@@ -0,0 +1,83 @@
+use anyhow::{anyhow, Result};
```
MGibson1, line 1: In general, all macos comments apply here as well.
* why async?
* why specified lifetimes?
* prefer `?` over `map_error`
* test needs work to not orphan secrets on failure

desktop_native/src/password/unix.rs
```diff
@@ -0,0 +1,83 @@
...
+use libsecret::{password_clear_sync, password_lookup_sync, password_store_sync, Schema};
```
MGibson1, line 2: Any chance we can use the `future` versions of these and actually put the `async` to use?

desktop_native/src/password/windows.rs
```diff
@@ -0,0 +1,158 @@
...
+        CredReadW(
+            PWSTR(target_name.as_ptr()),
+            CRED_TYPE_GENERIC.0,
+            CRED_FLAGS_NONE,
+            credential_ptr,
+        )
+    };
+
+    scopeguard::defer!({
+        unsafe { CredFree(credential as *mut _) };
```
MGibson1, line 29: It seems unlikely the crate is designed to be used with so much `unsafe`.

Is the issue that no wrapper functions exist for things like `CredFree` or `GetError`? I'm worried we're using the library wrong...

src/main.ts
```diff
@@ -2,19 +2,19 @@ import * as path from "path";
...
-import { KeytarStorageListener } from "jslib-electron/keytarStorageListener";
 import { ElectronLogService } from "jslib-electron/services/electronLog.service";
 import { ElectronMainMessagingService } from "jslib-electron/services/electronMainMessaging.service";
 import { ElectronStorageService } from "jslib-electron/services/electronStorage.service";
 import { TrayMain } from "jslib-electron/tray.main";
 import { UpdaterMain } from "jslib-electron/updater.main";
 import { WindowMain } from "jslib-electron/window.main";
 
-import { MenuMain } from "./main/menu.main";
+import { BiometricMain } from "./main/biometric/biometric.main";
```
MGibson1, line 15: let's revert all these biometric changes until we actually implement them in our native module

src/main/biometric/biometric.darwin.main.ts
```diff
@@ -0,0 +1,36 @@
+import { ipcMain, systemPreferences } from "electron";
```
MGibson1, line 1: move back into jslib

src/main/biometric/biometric.windows.main.ts
```diff
@@ -0,0 +1,146 @@
+import { ipcMain } from "electron";
```
MGibson1, line 1: move back into jslib

src/main/biometric/biometric.main.ts
```diff
@@ -0,0 +1,6 @@
+export abstract class BiometricMain {
```
MGibson1, line 1: move back into jslib

src/main/menu/menu.about.ts
```diff
@@ -4,7 +4,7 @@ import { I18nService } from "jslib-common/abstractions/i18n.service";
 import { UpdaterMain } from "jslib-electron/updater.main";
```
MGibson1, line 4: move `menubar.ts` too?

MGibson1 (changes requested):
We should set up a time to pair on the rust changes to decide on some style guidelines.

desktop_native/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+mod biometric;
+mod password;
+
+#[allow(dead_code)]
+#[napi]
+mod passwords {
+    /// Fetch the stored password from the keychain.
+    #[napi]
+    pub async fn get_password(service: String, account: String) -> napi::Result<String> {
+        super::password::get_password(service.as_str(), account.as_str())
```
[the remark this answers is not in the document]
Hinton: I like `&`, updated all.

desktop_native/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+            .await
+            .map_err(|e| napi::Error::from_reason(e.to_string()))
```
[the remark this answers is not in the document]
Hinton: It didn't really make sense to keep the debug info between languages. This would still give us an error code which should point to the root cause.

desktop_native/src/password/macos.rs
Hinton: Keeping the same pattern as windows? Or can you mix async and non async calls?

desktop_native/src/password/unix.rs
Hinton: The future apis are a bit complicated as they expect a main gio thread + a  loop. I should probably investigate this a bit further as some users have reported issues with the linux app locking when libsecrets can't communicate.

desktop_native/src/password/windows.rs
Hinton: Sadly it's supposed to be used that way. The samples all contain big chunks of `unsafe`.

src/main.ts
Hinton: I needed to move biometric and keytar logic from jslib to avoid having to update directory connector.

desktop_native/src/password/macos.rs
```diff
@@ -0,0 +1,43 @@
...
+pub async fn delete_password(service: &str, account: &str) -> Result<()> {
+    delete_generic_password(&service, &account).map_err(anyhow::Error::msg)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test() {
```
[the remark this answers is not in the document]
Hinton: @MGibson1 What do you think of this change?

desktop_native/src/password/macos.rs
Hinton: Removed async.

Hinton:
Note to self: Verify that all platforms behave the same when no password is found.

desktop_native/src/password/windows.rs
```diff
@@ -140,19 +140,14 @@ fn target_name(service: &str, account: &str) -> String {
 mod tests {
     use super::*;
 
-    #[tokio::test]
-    async fn test() {
-        set_password("BitwardenTest", "BitwardenTest", "Random")
-            .await
-            .unwrap();
+    #[test]
+    fn test() {
```
MGibson1, line 139: Can we get a test of `get_password_keytar` for the windows version?

README.md
```diff
@@ -16,7 +16,11 @@ The Bitwarden desktop app is written using Electron and Angular. The application
 
 - [Node.js](https://nodejs.org) v16.13.1 (LTS) or greater
 - NPM v8
-- Windows users: To compile the native node modules used in the app you will need the _Visual C++ toolset_, available through the standard Visual Studio installer. You will also need to install the _Microsoft Build Tools 2015_ and _Windows 10 SDK 17134_ as additional dependencies in the Visual Studio installer.
+- Rust (https://www.rust-lang.org/tools/install)
+- Windows:
+  - To compile the native node modules used in the app you will need the _Visual C++ toolset_, available through the standard Visual Studio installer. You will also need to install the _Microsoft Build Tools 2015_ and _Windows 10 SDK 17134_ as additional dependencies in the Visual Studio installer.
+- Unix:
```
MGibson1, line 22: Isn't this just linux? Mac has it's own dep

MGibson1 (changes requested):
Looks good on the whole, just two small requests

desktop_native/src/password/windows.rs
Hinton: That means we would need to implement the store method as well 😢

MGibson1: Actually I think it would be better to have a test dep on [keytar](https://crates.io/crates/keytar) that stores it for us and we can prove we read it properly.

That way, if Keytar ever updates, we'll fail

MGibson1 (changes requested):

MGibson1 (dismissed):

PR #1406, opened 2022-03-15
Disable disk caching in stateService

eliykat:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Fix https://app.asana.com/0/0/1201967392760891/f.

Steps to Reproduce
1. Open the Desktop app
2. Open Preferences and check "Minimize when copying to clipboard" 
3. Copy a vault item's field to the clipboard
4. Reopen the minimized app
5. Open Preferences and UN-check "Minimize when copying to clipboard" 
6. Copy a vault item's field to the clipboard

Expected Result
1. The app will no longer minimize after copying to clipboard

Actual Result
1. The app will still minimize after copying to clipboard
## Code changes

This is because desktop runs a separate stateService in the main process, which only has access to onDisk storage. But each stateService caches separately and doesn't communicate with the other, so they get out of sync.

Given the impending release I have disabled caching in both stateService instances for this client. A better fix can be prioritized for next cycle.

## Before you submit

### DIFF
--- a/src/app/services.module.ts
+++ b/src/app/services.module.ts
@@ -184,7 +184,8 @@ export function initFactory(
           secureStorageService,
           logService,
           stateMigrationService,
-          new StateFactory(GlobalState, Account)
+          new StateFactory(GlobalState, Account),
+          false // Do not use disk caching because this will get out of sync with the main process service
         ),
       deps: [
         StorageServiceAbstraction,
--- a/src/main.ts
+++ b/src/main.ts
@@ -83,7 +83,8 @@ export class Main {
       null,
       this.logService,
       null,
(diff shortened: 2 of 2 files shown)


addisonbeck:
@eliykat I pushed a change for this to turn back on caching in the render. I agree that having the services in perfect sync is ideal, but the render process needs a state cache to keep it from being a performance burden. The background process doesn't, as it doesn't receive as much traffic to and from state, and isn't responsible for ever saving anything to state (except OpenAtLogin, which is a special case that isn't effected here)

Hinton (approved):

PR #1432, opened 2022-03-23
PS-91 hide update menu option in mas builds

jlf0dev:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Our MAS build failed due to "Your app updates itself outside of the Mac App Store". We need to hide the update option from the Help menu if we build for MAS.

Fix will be cherry picked to `rc`

## Code changes

- **src/main/menu.about.ts:** hide update option if build is for MAS
- **src/main/menu.first.ts:** hide update option if build is for MAS

note: `isMacAppStore()` will return `undefined` if false. In the future we should revisit this to make sure it returns false, but I'm uncomfortable making that change in a hotfix for a release

## Before you submit


justindbaur (approved):
We should make sure we make a ticket for fixing the `isMacAppStore()` sooner rather than later.

PR #1456, opened 2022-03-25
Username generator

kspearrin:
## Type of change

- [ ] Bug fix
- [x] New feature development
- [ ] Other

## Objective

Add support for username generation.

Asana task: https://app.asana.com/0/1200578627114106/1201532987437763

## Code changes

- Add new translations
- Implement UI changes to accommodate username generation on the generator page (see screenshots below)
- Add generate username button to add/edit page

## Screenshots

image

image

image

## Testing requirements

1. Test existing password generation options from generator.
2. Test new username generation options from generator.
3. Test integration of username and password generation into item add/edit page.

## Before you submit

- [x] This change requires a **documentation update** (notify the documentation team)


src/app/accounts/settings.component.html
```diff
@@ -3,7 +3,9 @@
     <div class="modal-content">
       <div class="modal-body form">
         <div class="box">
-          <label class="settingsTitle">{{ "settingsTitle" | i18n: currentUserEmail }} </label>
+          <div class="box-header">
+            {{ "settingsTitle" | i18n: currentUserEmail }}
+          </div>
```
kspearrin, line 8: This style didn't match other modals. So I updated it to use box-header.

src/scss/buttons.scss
```diff
@@ -115,3 +115,57 @@
     }
   }
 }
+
+.icon-btn {
```
kspearrin, line 119: extracted `row-btn` out to a new re-usable style. `row-btn` now extends `icon-btn`.

src/scss/misc.scss
```diff
@@ -173,25 +174,44 @@ p.lead {
   }
 }
 
-.password-block {
+.modal-title {
```
kspearrin, line 177: same as box-header. now can use it outside a box.

src/app/vault/password-generator.component.html
```diff
@@ -62,148 +90,356 @@
...
+                type="button"
+                (click)="toggleOptions()"
+                appA11yTitle="{{ 'toggleVisibility' | i18n }}"
+              >
+                <i class="bwi bwi-plus-square" aria-hidden="true" [hidden]="showOptions"></i>
+                <i class="bwi bwi-minus-square" aria-hidden="true" [hidden]="!showOptions"></i>
+                {{ "options" | i18n }}
+              </button>
+            </div>
+            <div class="box-content condensed" [hidden]="!showOptions">
```
eliykat, line 293: Any reason to use `[hidden]` instead of `*ngIf` here (and below)?

src/app/vault/password-generator.component.html
```diff
@@ -62,148 +90,356 @@
...
+            </div>
+          </ng-container>
+        </ng-container>
+        <ng-container *ngIf="type === 'username'">
+          <div class="box">
+            <div class="box-header">
+              <button
+                type="button"
+                (click)="toggleOptions()"
+                appA11yTitle="{{ 'toggleVisibility' | i18n }}"
```
eliykat, line 286: Should expose state to a11y utilities:

```suggestion
                appA11yTitle="{{ 'toggleVisibility' | i18n }}"
                [attr.aria-expanded]="showOptions"
```

eliykat (changes requested):

src/app/accounts/settings.component.html
kspearrin: Maybe. I can check if that works. I think we use box-header in other areas for the "modal title". Seems like its used whenever there is an existing box. I added modal-title style when no existing boxes are already being used in that area of the modal.

src/app/vault/password-generator-history.component.html
```diff
@@ -10,7 +10,7 @@
             <div class="box-content-row box-content-row-flex" *ngFor="let h of history">
               <div class="row-main">
                 <div
-                  class="password-wrapper monospaced"
+                  class="generated-wrapper monospaced"
```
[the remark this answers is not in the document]
kspearrin: Good catch. I need to move that style back out on its own.

src/app/accounts/settings.component.html
kspearrin: I am reviewing these modal/section titles with product designers at the moment. Plan is to do a follow up PR to clean things up across the desktop app.

src/app/vault/password-generator.component.html
```diff
@@ -62,148 +90,356 @@
...
+              >
+                <i class="bwi bwi-plus-square" aria-hidden="true" [hidden]="showOptions"></i>
```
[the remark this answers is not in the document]
kspearrin: I am reviewing these modal/section titles with product designers at the moment. Plan is to do a follow up PR to clean things up across the desktop app.

src/app/vault/password-generator.component.html
```diff
@@ -4,56 +4,84 @@
...
-        <div class="box">
-          <div class="box-content condensed">
-            <a class="box-content-row" href="#" appStopClick appBlurClick (click)="regenerate()">
-              <i class="bwi bwi-fw bwi-generate" aria-hidden="true"></i>
-              {{ "regeneratePassword" | i18n }}
-            </a>
-            <a class="box-content-row" href="#" appStopClick appBlurClick (click)="copy()">
-              <i class="bwi bwi-fw bwi-clone" aria-hidden="true"></i> {{ "copyPassword" | i18n }}
-            </a>
+        <div class="generated-block" *ngIf="type === 'password'">
```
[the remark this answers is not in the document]
kspearrin: I initially tried, but couldn't figure out how to dynamically swap out `i18n` translations in the `appA11yTitle`. I tried `appA11yTitle="{{ type === 'password' ? 'copyPassword' : 'copyUsername' | i18n }}"` but that doesn't seem to work. Always stays with the value that renders initially. I see that we do conditionals like this in a few other places in the app and confirmed that those too are not working as expected.

src/app/vault/password-generator.component.html
kspearrin: Not sure I see a reason. Was just following existing code for password generator.

src/app/vault/password-generator.component.html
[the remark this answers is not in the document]
eliykat: I see, it looks like that directive only works on the initial input, it doesn't update on changes after that point. Might be worth changing that behaviour, but we can call it out of scope for this PR.

src/app/vault/password-generator.component.html
eliykat: Fair enough. I don't think it'll make a difference in this case, so we can leave it.

eliykat (approved):

Issue #1458, opened 2022-03-27
A11y: length range slider lacks visible focus indication and accessible name


patrickhlauke:
### Steps To Reproduce

For focus indication:

1. Navigating with the keyboard, go to "View > Password Generator", open "Options"
2. Set focus (using `Tab`/`Shift+Tab`) to the Length range slider

---

For accessible name:

1. Run a screenreader (e.g. NVDA on Windows)
2. go to "View > Password Generator", open "Options"
3. Set focus (using `Tab`/`Shift+Tab`) to the Length range slider

### Expected Result

For focus indication:

- expect to see a visible focus ring or indication on the range slider when it has focus

---

For accessible name:

- expect the slider to be announced with a specific label/name

### Actual Result

For focus indication:

- the range slider lacks any visible focus indication

---

For accessible name:

- the slider is simply announced as an unlabelled/generic "slider"

### Screenshots or Videos

For focus indication: animated gif showing part of the focus cycle in the password generation options. Note how when the range slider receives focus, there is no visible focus indication on the input itself.
![bitwarden-range-slider-focus](https://user-images.githubusercontent.com/895831/160306265-00dadcd7-c7b2-4429-9252-8ca5f44d7573.gif)

---

For accessible name: recording using NVDA on Windows, with Speech Viewer enabled. Note how the number/spinbox is announced correctly as "Length spin button ...", while the range slider itself is simply announced as an unlabelled "slider".

https://user-images.githubusercontent.com/895831/160306652-cd271edd-2bd5-4164-9132-87b2d9842a12.mp4

### Additional Context

I would actually suggest that the range slider should not be keyboard-focusable at all, as it is redundant for a keyboard user (since they can operate the preceding spin box/number input). This neatly sidesteps the issue of lack of visible focus indication.

For screen reader users though, I'd still suggest adding an explicit accessible name to the non-focusable range slider (thinking for instance of situations where a user may be partially sighted, use a mouse, but also a screen reader ... they'll want to hear the accessible name of the control once they're operating it/it's under their mouse). A simple `aria-label`, replicating the "Length" label, should be sufficient.

### Operating System

Windows

### Operating System Version

10

### Installation method

Direct Download (from bitwarden.com)

### Build Version

Version 1.32.1 Shell 16.0.7 Renderer 96.0.4664.110 Node 16.9.1 Architecture x64

PR #1459, opened 2022-03-28
[PS-120] Add `aria-label` to password generator length slider, make it non-keyboard-focusable

patrickhlauke:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

Currently the range slider in the password generator options receives keyboard focus when navigating through the modal using `Tab`/`Shift+Tab`, but it lacks a focus indicator. Fundamentally though, the slider is redundant for keyboard users, as they can already operate the number input/spin box that precedes it. This PR removes the range slider from the keyboard focus cycle using `tabindex="-1"`.

Further, the range slider currently lacks an accessible name. Assistive technologies just announce it as a generic unlabelled "slider". This PR adds an explicit `aria-label` that mirrors the label for the preceding number input. This way, assistive tech users that still manage to reach the slider (e.g. using reading keys, rather than `Tab`/`Shift+Tab`) will still hear it announced sensibly/correctly.

Closes https://github.com/bitwarden/desktop/issues/1458

## Screenshot

Video recording using NVDA with Speech Viewer. Current state, showing that the range slider receives focus (and shows no visible focus indication) when navigating using `Tab`/`Shift+Tab` and is announced as generic unlabelled "slider". Then the same dialog after this PR is applied - note that navigating using `Tab`/`Shift+Tab`, the range slider doesn't receive focus at all (focus goes from the length number input directly to the A-Z checkbox). Changing NVDA to get out of forms mode (`NVDA+Space`) and navigating using reading keys (cursor keys), the range slider can be reached, and is announced as "Length slider...".

https://user-images.githubusercontent.com/895831/160355562-7f1140a9-06dc-4d9e-b724-85fb2fe9b7fc.mp4

## Testing requirements

- test using keyboard navigation first (noting the slider is now not in the focus cycle, so it doesn't receive focus and then disorient users by not actually having a focus indicator)
- test using a screen reader (e.g. NVDA on Windows) - using reading keys (generally, cursor keys), reach the slider and note that it's announced not just as "slider" but as "Length slider"

## Before you submit


eliykat (approved):
Thanks @patrickhlauke, really appreciate your work on accessibility.

patrickhlauke:
@eliykat no worries, my pleasure. particularly for relatively easy/low-hanging fruit like this (that don't need fundamental reworking of logic or a deep knowledge of angular, which I lack), I'm always happy to help 👍

PR #1473, opened 2022-04-04
[bug] Add dev overrides to the renderer's webpack config

addisonbeck:
https://bitwarden.atlassian.net/jira/software/projects/SG/boards/34?selectedIssue=SG-178

## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective
Errors from the renderer are minified right now. This PR adds a dev webpack config that handles this conditionally based on environment. It follows patterns for this already present in the main process's webpack config.

## Code changes
* Added a dev specific webpack config to the renderer with instructions not to minimize.

### DIFF
--- a/webpack.renderer.js
+++ b/webpack.renderer.js
@@ -41,7 +41,7 @@ const common = {
 };
 
 const renderer = {
-  mode: "production",
+  mode: NODE_ENV,
   devtool: "source-map",
   target: "electron-renderer",
   node: {


webpack.renderer.js
```diff
@@ -40,16 +40,13 @@ const common = {
...
-  mode: "production",
-  devtool: "source-map",
-  target: "electron-renderer",
-  node: {
-    __dirname: false,
-  },
-  entry: {
-    "app/main": "./src/app/main.ts",
+const dev = {
+  optimization: {
```
Hinton, line 44: Do we need this? During my testing it was sufficient to just set mode.

addisonbeck: You're right, that was all that was needed. Fixed.

Hinton (approved):

PR #1496, opened 2022-04-22
[EC-156] [BEEEP] Remove factory providers in Angular DI

eliykat:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

See https://github.com/bitwarden/jslib/pull/775.
## Code changes

* update services.module.ts to use the new tokens
* extract init logic to an initService so that it's not mixed in with provider logic

## Before you submit


src/app/services/init.service.ts
```diff
@@ -0,0 +1,81 @@
...
+      (this.vaultTimeoutService as VaultTimeoutService).init(true);
+      const locale = await this.stateService.getLocale();
+      await (this.i18nService as I18nService).init(locale);
+      (this.eventService as EventService).init(true);
+      this.twoFactorService.init();
+      setTimeout(() => this.notificationsService.init(), 3000);
+      const htmlEl = this.win.document.documentElement;
+      htmlEl.classList.add("os_" + this.platformUtilsService.getDeviceString());
+      htmlEl.classList.add("locale_" + this.i18nService.translationLocale);
+      const theme = await this.platformUtilsService.getEffectiveTheme();
```
Hinton, line 54:
```suggestion

      const theme = await this.platformUtilsService.getEffectiveTheme();
```

src/app/services/init.service.ts
```diff
@@ -0,0 +1,81 @@
...
+      this.syncService.fullSync(true);
+      (this.vaultTimeoutService as VaultTimeoutService).init(true);
+      const locale = await this.stateService.getLocale();
+      await (this.i18nService as I18nService).init(locale);
+      (this.eventService as EventService).init(true);
+      this.twoFactorService.init();
+      setTimeout(() => this.notificationsService.init(), 3000);
+      const htmlEl = this.win.document.documentElement;
+      htmlEl.classList.add("os_" + this.platformUtilsService.getDeviceString());
+      htmlEl.classList.add("locale_" + this.i18nService.translationLocale);
```
Hinton, line 53: Do you know why we are setting a class for locale?

src/app/services/services.module.ts
```diff
@@ -0,0 +1,126 @@
...
+    },
+    {
+      provide: SystemServiceAbstraction,
+      useFactory: (
+        messagingService: MessagingServiceAbstraction,
+        platformUtilsService: PlatformUtilsServiceAbstraction,
+        stateService: StateServiceAbstraction
+      ) => new SystemService(messagingService, platformUtilsService, null, stateService),
+      deps: [MessagingServiceAbstraction, PlatformUtilsServiceAbstraction, StateServiceAbstraction],
+    },
```
Hinton, line 110: Did we forget to change this from `useFactory`?

src/app/services/services.module.ts
```diff
@@ -0,0 +1,126 @@
...
+      deps: [
+        I18nServiceAbstraction,
+        MessagingServiceAbstraction,
+        CLIENT_TYPE,
+        StateServiceAbstraction,
+      ],
+    },
+    {
+      provide: I18nServiceAbstraction,
+      useFactory: (window: Window) => new I18nService(window.navigator.language, "./locales"),
```
Hinton, line 82: Could we use a variable for `locales`?

eliykat: I assume you mean token - done, depends on https://github.com/bitwarden/jslib/pull/783

src/app/services/services.module.ts
eliykat: Missed this one - done. Put the token declaration in this file (not jslib) because it's not used by any other `services.module.ts`. (only in browser, and that's in the background)

src/app/services/init.service.ts
eliykat: I don't, but here's the original git blame: https://github.com/bitwarden/desktop/commit/5fdfce0bff12a9dd594c702add7a26848d749d2b

It looks like it was added at the same time as the `os_` class, in case locale-specific styling was required. I can't see that we use it anywhere though, so I think it's safe to remove. (and I've removed it)

src/app/services/services.module.ts
Hinton: Yep. What do you think of also doing the same for `window.navigator.language`? Would let us remove the factory here completely.

eliykat: I didn't know whether it would be going overboard, but I'm happy to :D this is done in the jslib PR and will be added here shortly

Hinton (approved):

Issue #1332, opened 2022-02-13
#a11y: Lack of clearly visible focus indication for some interface components


patrickhlauke:
### Steps To Reproduce

1. Open Bitwarden
2. Use `Tab` / `Shift+Tab` to navigate using the keyboard

### Expected Result

A visible keyboard focus indication on the "My Vault" and "Send" controls.

### Actual Result

While there does appear to be a subtle shift in background colour, there's no clear indication which one has focus (compare with other elements that have a reasonably clear indicator)

### Screenshots or Videos

Animated GIF of navigating through the application - note that after the "No folder" focus goes to "My Vault", then "Send", but there really isn't much of a visual indication.
![bitwarden-desktop-focus-indication](https://user-images.githubusercontent.com/895831/153774875-0ea38f4d-b564-4e1e-a069-ed762ff9ac76.gif)

### Operating System

Windows

### Operating System Version

10

### Installation method

Direct Download (from bitwarden.com)

### Build Version

Version 1.31.2 Shell 16.0.7 Renderer 96.0.4664.110 Node 16.9.1 Architecture x64

Hinton:
Hi @patrickhlauke,

Thanks for reporting this. If I understand it correctly it's primarily the `My Vault` and `Send` tabs in the bottom left corner that are missing focus styling? The `Test Identity` seems to also lack the top border as it's obstructed by the search bar.

patrickhlauke:
yes, primarily the `My Vault` and `Send` since they pretty much lack any discernible/clear focus indication. the `Test Identity` (and well, all items really) could do with a proper full outline (not just the little sliver at the top) if feasible, but at least they arguably have something visible already there (with the subtle change in colour and the left-hand border changing)

squarepupil:
I will also note that I only discovered this issue because the UI behavior recently changed when hitting tab after entering a search query. It used to jump the cursor to the list of search results and now (frustratingly) jumps the cursor to the account selection dropdown followed by the menu on the left side of the application, which can be seen in the gif at the top of this issue report.

As a by-the-way, does anyone know how to quickly get to your first search result without having to touch the mouse?

patrickhlauke:
@squarepupil this will likely be one of those "squaring the circle" type problems. for some folks the old focus order will make sense, for others the current one. i think the only way around that would be to start treating the different areas/panels in the app as separate regions that you jump between using something like `F6` (standard at least on Windows, can't remember offhand if macOS and Linux have a similar key)

[...]

PR #1514, opened 2022-05-01
[PS-540] Accessibility: remove `appBlurClick`, add `aria-expanded` to cog options, links to buttons, fix unsufficient focus indication

patrickhlauke:
## Type of change

- [x] Bug fix
- [ ] New feature development
- [ ] Other

## Objective

* "My Vault" / "Send" navigation components act as toggle buttons, so turning them to `<button>` elements and adding `aria-pressed` to convey their state (currently, they're only visually distinguished with the `active` class)
* In the add/edit view, add `aria-expanded` to the "cog" options controls (for adding new URIs)
* Change any straggler `<a>` links to buttons where appropriate (where they don't open a browser window)
* Currently, for many of the controls, activating the controls leads to focus being lost/reset to the start of the document. This is cause by the `blur()` that happens due to the `appBlurClick`. This can cause problems for screen reader users (in particular, with JAWS) as it resets the reading cursor to the start of the document. There should be no need to `blur()` things (assuming this was originally done to remove the visual focus indicator, but these days this is handled automatically by chromium, not showing the outline when things are clicked with a mouse/tapped on a touchscreen). This PR removes all instances of `appBlurClick` as they're really not needed, and *not* `blur()`ring things fixes a few inconsistencies/problems for keyboard and screen reader users.
* reenable focus outline for all buttons - this fixes the main problem of "My Vault" / "Send" not having clear focus, and makes focus indication on other controls clearer/more consistent. See https://github.com/bitwarden/desktop/issues/1332

Closes https://github.com/bitwarden/desktop/issues/1332

## Code changes

- **src/app/vault/add-edit.component.html**: add `aria-expanded` to the cog-icon options
- **src/scss/vault.scss**:  remove the unnecessary 10px margin on buttons in the vault footer (this caused incorrect display once "My Vault" / "Send" were changed to `<button>`
- **src/scss/button.scss**: remove the `outline:0` focus outline suppression. Fixes the lack of clear focus indication on "My Vault" / "Send" and makes other buttons more consistent/clear in their keyboard outline

All other files changed to essentially remove `appBlurClick` which is not needed/causes problems. Now nothing in the desktop app uses this directive at all.

## Screenshots/videos

Before/after video (using NVDA) to show that currently, "My Vault" / "Send" are exposed as links, and they do not expose programmatically which one is active/selected. At 0:12, it shows the fix in this PR - exposing them as buttons, with appropriate `aria-pressed` to indicate which one is/isn't active.

https://user-images.githubusercontent.com/895831/166142378-61bb89d0-36fe-4f5e-933f-1a66b6b5239d.mp4

---

Before/after video of the "cog" options. Currently, the state of the options is not announced. After this change (0:05), the cog announces whether it's expanded or collapsed.

https://user-images.githubusercontent.com/895831/166142436-49d9a086-ecac-4b50-bdf7-cae7fad993d1.mp4

---

Long video showing the effect of removing `appBlurClick`.

https://user-images.githubusercontent.com/895831/166142529-d6977c5a-d787-4b36-a851-a86cf3c1015b.mp4

* 0:00 - 0:17 current behavior of "Options" expand/collapse in add/edit view with appBlurClick - note in the devtools console how `document.activeElement` keeps resetting to body element. browsers/engines do try to error-correct for this by then finding the nearest focusable element to where you previously were when using `Tab`/`Shift+Tab`, but this behavior is unreliable (see later)
* 0:17 - 0:26 after removing appBlurClick, focus remains on the button that was activated. This also results in screen reader dynamically announcing the change from collapsed to expanded.
* 0:26 - 0:40 current behavior of filter toggle buttons (in My Vault, but same in Send) with appBlurClick - note again in devtools how `document.activeElement` resets to body, and the browser engine error correction for `Tab`bing forward.
* 0:40 - 0:54 after removing appBlurClick, focus remains on the filter button that was activated. This also results in screen reader dynamically announcing the change from not pressed to pressed.
* 0:54 - 1:18 using JAWS screen reader, demonstration of what currently happens when focus is lost (with filters, but applicable to all other cases where focus goes back to body). The document's title ("Bitwarden document") is announced again, and the reading/virtual cursor for JAWS is reset to the very start of the document - using reading/cursor keys, note how we're now right back at the top with the search box.
* 1:18 - 1:38 after removing appBlurClick, the document title ("Bitwarden document") is not re-announced. Using reading/cursor keys (1:35) shows that reading/virtual cursor remained on the just toggled button and reading carries on from that point, rather than from the very top of the document

---

Animated gif showing the fixed/improved focus indication when navigating through the app using the keyboard - compare this to the gif in https://github.com/bitwarden/desktop/issues/1332

animated gif showing focus outline when navigating through the application

## Testing requirements

- test using a screen reader and keyboard

## Before you submit

I got one linting warning

```
.../src/app/services/services.module.ts
  8:3  warning  'WINDOW' is defined but never used  @typescript-eslint/no-unused-vars
```

but as this is not related to any changes I made, assuming this is something to be fixed in `master`


Hinton:
I believe the `blur` on submit buttons is to remove the focus state. @danielleflinn do you have any inputs on this design-wise?

patrickhlauke:
if it's to remove focus state, there's modern alternatives like using `:focus-visible` instead of `:focus`

Hinton:
I believe the button styling got mixed up?
image

Hinton:
@patrickhlauke there are scenarios where buttons stays around after clicking, like attempting to save an item when offline. I'm not arguing that keeping blur is correct just that it introduces visual differences that we will need to consider.

src/scss/vault.scss
```diff
@@ -414,14 +414,6 @@ app-root {
...
     }
 
-    button {
-      margin-right: 10px;
-
-      &:last-child {
-        margin-right: 0;
-      }
-    }
-
```
Hinton, line 424: This is used for buttons when editing an item, and we'll need to keep it for now.

Hinton (commented):
I believe this looks good, the buttons when modifying an item got messed up. And we'll need to wait for design input on the changed buttons. One alternative is to look into using `:focus-visible` instead which should keep the visual while offering better accessibility.

src/scss/vault.scss
patrickhlauke: oh, I missed that. I can have a second look at this later tonight.

patrickhlauke:
> I believe the button styling got mixed up? image

I'll review this tonight, must have slipped me by

src/scss/vault.scss
patrickhlauke: ah i know what happened there. removed that margin to fix the "My Vault"/"Sends" (now changed to buttons) having a gap, but missed that this was needed in the footer of the edit view for items.

patrickhlauke:
> @patrickhlauke there are scenarios where buttons stays around after clicking, like attempting to save an item when offline. I'm not arguing that keeping blur is correct just that it introduces visual differences that we will need to consider.

no worries, i didn't interpret this as arguing for blur, so all good. was more surprised as i hadn't come across buttons that remain visible after submission. i'll have a look/see if i can replicate this and propose a tweak

patrickhlauke:
@Hinton i reintroduced the button margin styles there, but more specifically just for when the `.footer` is in `.details`.

footers across the application, with margins only applied in the footer of the details view

details edit view buttons now with correct margin

Hinton (approved):

Hinton:
This looks good, thanks for yet another contribution @patrickhlauke.

patrickhlauke:
> This looks good, thanks for yet another contribution @patrickhlauke.

working towards the "community contributor of the century" award ;)

but no seriously, my pleasure. it's nice to see the products improving bit by bit, and knowing i played a small part in it. more to come, no doubt...

PR #1521, opened 2022-05-02
Update publish settings to use S3

mimartin12:
## Type of change

- [ ] Bug fix
- [ ] New feature development
- [X] Build/deploy pipeline (DevOps)
- [ ] Other

## Objective

Allows the application to look for the latest updates on S3, instead of Github.

## Code changes

- **.github/workflows/release.yml:** Now uploads the artifacts and electron update files to a S3 bucket.
- **electron-builder.json:** Adds S3 publish configuration so that the electron app looks at the S3 bucket for updates.

## Before you submit

- [X] This change has particular **deployment requirements** (notify the DevOps team)


electron-builder.json
```diff
@@ -8,6 +8,13 @@
   "afterSign": "scripts/after-sign.js",
   "asarUnpack": ["**/*.node"],
   "files": ["**/*", "!**/node_modules/@bitwarden/desktop-native/**/*"],
+  "publish": {
```
Hinton, line 11: We need to define `endpoint` here https://www.electron.build/configuration/publish, otherwise it will attempt to fetch it directly from S3 which we don't want.

mimartin12: Hey thanks for the reminder. I just missed that one 👍 .

.github/workflows/release.yml
```diff
@@ -150,9 +171,6 @@ jobs:
           artifacts: bitwarden_${{ env._PKG_VERSION }}_amd64.snap
           path: ./dist
```
vgrassia, line 172: Shouldn't `artifacts` stay the same and the `path` key be changed to `./artifacts`?

vgrassia (changes requested):

vgrassia (approved):

